PR #7, opened 2023-12-18
add EXTERNALLY-MANAGED

### LINKED ISSUE #1: Install EXTERNALLY-MANAGED
Put this file in its expected place with a message that says:

- This environment is protected against naive pip installs
- To use 'conda pip' instead for extra safety, or `--break-system-packages`
- Where to find more information about the rationale

jaimergp:
Closes #1 

In this PR, we are implementing the following behavior:

- `conda-pip` packaged files include `EXTERNALLY-MANAGED` by default. This protects the `base` environment (or whatever `conda-pip` has been installed to; usually `base`).
- On target environments, we will place `EXTERNALLY-MANAGED` after `conda pip -p target install ...` executions.
- Thanks to a post-command plugin:
  - We will also inject `EXTERNALLY-MANAGED` _if_ the target environment has `pip`.
  - In Python updates, the plugin will also migrate the file from one `lib/pythonX.Y` directory to the new one.
  - We will _remove_ the marker file when `pip` is not present in the target environment anymore.
  - This is not done in envs created by conda-build, or if `conda-pip` is present in `base` as a dependency and not explicitly installed (we check the history file for explicit presence of `conda-pip`).

To do / discuss:
- What happens when `conda remove --all` is invoked? This should be caught by the post-command plugin anyway, but a few empty directories might be in place? Maybe `conda` clears the tree? Investigate.
- Same but for `conda env remove`.

References:

- https://peps.python.org/pep-0668/
- https://packaging.python.org/en/latest/specifications/externally-managed-environments
- https://docs.python.org/3/library/sysconfig.html#sysconfig.get_paths


jaimergp:
I'm not too sure yet about writing the files directly to the target environment. It creates the opportunity for lingering files unless the user removes packages nicely. Maybe an alternative is to inject a fake `conda-meta/externally-managed-1.0-py3xy.json` metadata file that lists the `EXTERNALLY-MANAGED` file so it can be `conda remove`d nicely. I'll leave that for a further issue.

jaimergp:
Pinging @jezdez @pradyunsg for awareness / comments if time allows :)

conda_pip/data/EXTERNALLY-MANAGED
```diff
@@ -0,0 +1,11 @@
...
+Error=You are trying to modify a conda environment!
+  
+  If you want to install PyPI packages into this conda environment,
+  please consider using 'conda pip install ...' instead. 
+  This will make a few checks to ensure that the PyPI packages 
+  you want to install will not disrupt your conda environment.
+
+  If you still want to proceed with 'pip install ...', then
+  add the '--break-system-packages' to your command, but be aware
+  that you might experience unexpected issues.
```
jezdez, line 11: `PIP_BREAK_SYSTEM_PACKAGES` or the config option might also be worth mentioning. Or linking to the pip docs about it?

conda_pip/main.py
```diff
@@ -110,13 +125,71 @@ def run_pip_install(
...
+    prefix_data = PrefixData(target_prefix)
+    if command in {"create", "install", "update"}:
+        # ensure target env has pip installed
+        if not list(prefix_data.query("pip")):
+            return
+        # Check if there are some leftover EXTERNALLY-MANAGED files from other Python versions
+        if command != "create" and os.name != "nt":
+            for python_dir in Path(target_prefix, "lib").glob("python*"):
+                if python_dir.is_dir():
+                    externally_managed = Path(python_dir, "EXTERNALLY-MANAGED")
```
jezdez, line 176:
```suggestion
                    externally_managed = python_dir / "EXTERNALLY-MANAGED"
```

conda_pip/main.py
```diff
@@ -110,13 +125,71 @@ def run_pip_install(
...
+    conda-pip places its own EXTERNALLY-MANAGED file when it is installed in an environment.
+    We also need to place it in _new_ environments created by conda. We do this by implementing
+    some extra plugin hooks.
+    """
+    base_dir = get_env_stdlib(prefix or sys.prefix)
+    externally_managed = Path(base_dir, "EXTERNALLY-MANAGED")
+    if externally_managed.exists():
+        return
+    logger.info("Placing EXTERNALLY-MANAGED in %s", base_dir)
+    shutil.copy(HERE / "data" / "EXTERNALLY-MANAGED", externally_managed)
```
jezdez, line 148: This should use https://docs.python.org/3/library/importlib.resources.html instead of `__file__` for portability

conda_pip/main.py
```diff
@@ -110,13 +125,71 @@ def run_pip_install(
...
+        # Check if there are some leftover EXTERNALLY-MANAGED files from other Python versions
+        if command != "create" and os.name != "nt":
+            for python_dir in Path(target_prefix, "lib").glob("python*"):
+                if python_dir.is_dir():
+                    externally_managed = Path(python_dir, "EXTERNALLY-MANAGED")
+                    if externally_managed.exists():
+                        externally_managed.unlink()
+        place_externally_managed(target_prefix)
+    else:  # remove
+        if list(prefix_data.query("pip")):
```
jezdez, line 181: Any exception expected with this call?

conda_pip/main.py
```diff
@@ -110,13 +125,71 @@ def run_pip_install(
...
+            # leave in place if pip is still installed
+            return
+        if os.name == "nt":
+            externally_managed = Path(target_prefix, "Lib", "EXTERNALLY-MANAGED")
```
jezdez, line 185: Let's factor out getting the path to an EXTERNALLY-MANAGED file, might be useful elsewhere and would be good to test

conda_pip/plugin.py
```diff
@@ -14,3 +14,12 @@ def conda_subcommands():
...
         configure_parser=configure_parser,
     )
+
+
+@plugins.hookimpl
+def conda_post_commands():
+    yield plugins.CondaPostCommand(
+        name="conda-pip-ensure-target-env-has-externally-managed",
+        action=ensure_target_env_has_externally_managed,
+        run_for={"install", "create", "update", "remove"},
```
jezdez, line 24: This should also cover the various `env` subcommands

jezdez (commented):
Noice!

conda_pip/data/EXTERNALLY-MANAGED
jaimergp: Adding both in the next push.

conda_pip/plugin.py
jaimergp: It's tricky to know how we should handle this in combination with the potential `conda env` hook. I'd rather think about in a different PR so the conversations are well captured separately.

conda_pip/main.py
jaimergp: `.query()` returns a generator. Worst case it's empty because there were no matches.

conda_pip/main.py
jaimergp: I like this suggestion, thanks!

PR #8, opened 2023-12-18
add tests

jaimergp:


jaimergp:
There's an installation issue on Windows that affects `conda`'s ability load the `libmamba` libraries for the solver:

```
WARNING  conda.plugins.manager:manager.py:134 Error while loading conda entry point: conda-libmamba-solver (The specified procedure could not be found)
```

Found this issue: https://github.com/mamba-org/mamba/issues/1728, but it's not very useful. This is normally a result of mismatched ABIs when mixing channels, but we are only using conda-forge here 🤷 I also tried downgrading to see if it was a 1.5.6 issue, but same error. I'll investigate in a subsequent PR. Disabling for now.

PR #23, opened 2024-05-09
Implement explicit lockfiles integration

### LINKED ISSUE #21: Integrate with explicit lockfiles
`conda list --explicit --md5` will produce same-platform lockfiles that can be installed somewhere else via `conda install --file ...`. These are good enough for many cases, but it doesn't handle PyPI dependencies.

I suggest:

- We add a post_command plugin for `list` and add extra lines with `# pypi: url-to-wheel`.
- Add a post-command hook to post process just the input files (not the CLI specs) and look for `# pypi: url` lines.

jaimergp:
Closes #21 

- [X] `conda list --explicit` extensions
- [x] `conda install | create` extensions for the above output
- [x] Tests
- [x] Docs


jaimergp:
This is how the output of `conda list --explicit` would work, btw:

```
# This file may be used to create an environment using:
# $ conda create --name <env> --file <this file>
# platform: osx-arm64
@EXPLICIT
https://conda.anaconda.org/conda-forge/osx-arm64/bzip2-1.0.8-h93a5062_5.conda
https://conda.anaconda.org/conda-forge/osx-arm64/c-ares-1.23.0-h93a5062_0.conda
https://conda.anaconda.org/conda-forge/osx-arm64/ca-certificates-2024.2.2-hf0a4a13_0.conda
https://conda.anaconda.org/conda-forge/osx-arm64/icu-73.2-hc8870d7_0.conda
https://conda.anaconda.org/conda-forge/osx-arm64/jpeg-9e-h1a8c8d9_3.conda
https://conda.anaconda.org/conda-forge/osx-arm64/libcxx-16.0.6-h4653b0c_0.conda
https://conda.anaconda.org/conda-forge/osx-arm64/libdeflate-1.17-h1a8c8d9_0.conda
https://conda.anaconda.org/conda-forge/osx-arm64/libedit-3.1.20191231-hc8eb9b7_2.tar.bz2
https://conda.anaconda.org/conda-forge/osx-arm64/libev-4.33-h93a5062_2.conda
https://conda.anaconda.org/conda-forge/osx-arm64/libffi-3.4.2-h3422bc3_5.tar.bz2
https://conda.anaconda.org/conda-forge/osx-arm64/libiconv-1.17-h0d3ecfb_1.conda
https://conda.anaconda.org/conda-forge/osx-arm64/libwebp-base-1.3.2-hb547adb_0.conda
https://conda.anaconda.org/conda-forge/osx-arm64/libzlib-1.2.13-h53f4e23_5.conda
https://conda.anaconda.org/conda-forge/osx-arm64/llvm-openmp-17.0.6-hcd81f8e_0.conda
https://conda.anaconda.org/conda-forge/osx-arm64/lzo-2.10-h642e427_1000.tar.bz2
# Following lines added by conda-pypi
# pypi: gitpython==3.1.32 --python-version 3.9.9 --implementation cp
# pypi: platformdirs==4.2.2 --python-version 3.9.9 --implementation cp
# pypi: sphinx==4.2.0 --python-version 3.9.9 --implementation cp
# pypi: python-dateutil==2.8.1 --python-version 3.9.9 --implementation cp
# pypi: pathspec==0.8.1 --python-version 3.9.9 --implementation cp
# pypi: charset-normalizer==2.0.7 --python-version 3.9.9 --implementation cp
# pypi: tornado==6.1 --python-version 3.9.9 --implementation cp --abi cp39 --platform macosx_11_0_arm64
# pypi: gitdb==4.0.10 --python-version 3.9.9 --implementation cp
# pypi: black==21.6b0 --python-version 3.9.9 --implementation cp
# pypi: regex==2021.4.4 --python-version 3.9.9 --implementation cp --abi cp39 --platform macosx_11_0_arm64
# pypi: tzdata==2023.3 --python-version 3.9.9 --implementation cp
# pypi: pillow==8.3.2 --python-version 3.9.9 --implementation cp --abi cp39 --platform macosx_11_0_arm64
```

We can't print the URL for PyPI wheels because it wasn't kept around 😬

jaimergp:
I'm not personally a fan of using the CLI flags here but these are the problems I faced trying to obtain something else:

- There's no record of the origin URL for an installed PyPI package (see https://github.com/pypa/pip/issues/10736). This might change with [PEP-710](https://peps.python.org/pep-0710/), but it's still a draft.
- The environment markers (`name=version; python_version='2.7' etc) do not limit which wheels will be obtained, but which lines are accepted for inclusion in the request list given the current Python interpreter.

I wish there was a way to encode the constraints in a single string field (like a URI or something), but right now this is the only way, I think.

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+        ignore = False
+        wheel = {}
+        hashed_record = ""
+        for path in record.files:
+            path = Path(context.target_prefix, path)
+            if "__editable__" in path.stem:
+                ignore = True
+                break
+            if path.name == "direct_url.json" and path.parent.suffix == ".dist-info":
+                data = json.loads(path.read_text())
```
jezdez, line 226: this could raise OSError with permissions errors etc and should be handled more defensively

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+            lines[-1] += f" --implementation {python_details['implementation']}"
+            if wheel and (wheel_tag := wheel.get("Tag")):
+                for tag in wheel_tag:
+                    _, abi_tag, platform_tag = tag.split("-", 2)
+                    if abi_tag != "none" and abi_tag not in seen["abi"]:
+                        lines[-1] += f" --abi {abi_tag}"
+                        seen["abi"].add(abi_tag)
+                    if platform_tag != "any" and platform_tag not in seen["platform"]:
+                        lines[-1] += f" --platform {platform_tag}"
+                        seen["platform"].add(platform_tag)
```
jezdez, line 259: Oh no, manually parsing tag info :( This must be available in packaging somewhere

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+            # Here we could try to run a --dry-run --report some.json to get the resolved URL
+            # but it's not guaranteed we get the exact same source so for now we defer to install
+            # time
+        if checksum and hashed_record:
+            lines[-1] += f" -- --record-checksum={checksum}:{hashed_record}"
+
+    return lines
+
+
+def dry_run_pip_json(args: Iterable[str], force_reinstall: bool = False) -> dict[str, Any]:
```
jezdez, line 269: needs docs

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+    # pip can output to stdout via `--report -` (dash), but this
+    # creates issues on Windows due to undecodable characters on some
+    # project descriptions (e.g. charset-normalizer, amusingly), which
+    # makes pip crash internally. Probably a bug on their end.
```
jezdez, line 273: let's file a bug upstream

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+    json_output = NamedTemporaryFile(suffix=".json", delete=False)
+    json_output.close()  # Prevent access errors on Windows
+
+    cmd = [
+        sys.executable,
+        "-mpip",
+        "install",
+        "--dry-run",
+        "--ignore-installed",
+        *(("--force-reinstall",) if force_reinstall else ()),
```
jezdez, line 284: what's with the unpacking?

```suggestion
        "--force-reinstall" if force_reinstall else "",
```

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+        "--report",
+        json_output.name,
+        "--target",
+        json_output.name + ".dir",
```
jezdez, line 288: why .dir?

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+            f"  stderr:\n{process.stderr}\n"
+            f"  stdout:\n{process.stdout}"
+        )
+
+    with open(json_output.name, "rb") as f:
+        # We need binary mode because the JSON output might
+        # contain weird unicode stuff (as part of the project
+        # description or README).
+        report = json.loads(f.read())
+    os.unlink(json_output.name)
```
jezdez, line 306: this should be done in a big try/finally block to make sure this file is deleted no matter what.

Also the `*.dir` directory created by the --target option above needs cleaning up

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+    return report
+
+
+def compute_record_sum(record_path, algo):
+    record = Path(record_path).read_text()
```
jezdez, line 311: OSError again

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+    lines = []
+    for line in record.splitlines():
+        path, *_ = line.split(",")
+        path = Path(path)
+        if path.parts[0].endswith(".dist-info") and path.name not in ("METADATA", "WHEEL"):
+            # we only want to check the metadata and wheel parts of dist-info; everything else
+            # is not deterministic or useful
+            continue
+        if path.parts[0] == ".." and any(part in path.parts for part in ("bin", "lib", "Scripts")):
```
jezdez, line 320:
```suggestion
        elif path.parts[0] == ".." and any(part in path.parts for part in ("bin", "lib", "Scripts")):
```

jezdez (commented):
A few notes:
- I'm not sure how I feel you expanding this format without a spec and a CEP since that's definitely a core functionality in conda
- There has to be other ways than implementing custom argv parsing and custom Python packaging standards parsing (big maintenance risk based on my experience)
- Inline docs and high-level reasoning for some of the design decisions are sorely needed before merging this

jezdez (commented):
A few notes:
- I'm not sure how I feel you expanding this format without a spec and a CEP since that's definitely a core functionality in conda
- There has to be other ways than implementing custom argv parsing and custom Python packaging standards parsing (big maintenance risk based on my experience)
- Inline docs and high-level reasoning for some of the design decisions are sorely needed before merging this

jaimergp:
Thanks for the thorough review!

> I'm not sure how I feel you expanding this format without a spec and a CEP since that's definitely a core functionality in conda

I do want the CEP. For the original format too. There's no standard to refer to other than the implementation. It should be straightforward.

> There has to be other ways than implementing custom argv parsing and custom Python packaging standards parsing (big maintenance risk based on my experience)

`conda` has [its own parsers](https://github.com/conda/conda/blob/main/conda/common/pkg_formats/python.py) but I only learnt that after writing this prototype. I'm sure we can find some parsers somewhere in `packaging` and friends too. Right now I'm mostly exploring how this idea would shape up, but I agree we need to polish this custom code better. Even the structure of the code could use some improvement.

> Inline docs and high-level reasoning for some of the design decisions are sorely needed before merging this

Yep, just wanted to get some external confirmation that this idea is worth pursuing before writing details on something that might not fly. I take that this review is also a 👍 to proceed further?

conda_pypi/main.py
jaimergp: I don't want to pass an empty string which subprocess will quote as a positional arg, won't it?

conda_pypi/cli/install.py
```diff
@@ -0,0 +1,111 @@
...
+from conda.base.context import context
+from conda.common.io import Spinner
+
+from ..main import run_pip_install, dry_run_pip_json, compute_record_sum
+from ..utils import get_env_site_packages
+
+if TYPE_CHECKING:
+    from typing import Iterable
+
+log = getLogger(f"conda.{__name__}")
```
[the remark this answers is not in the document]
jaimergp: I've been doing this in all my plugins to reuse `conda`'s configuration for consistency. CLS does this too, for example.

conda_pypi/cli/list.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def post_command(command: str):
+    if command != "list":
+        return
+    cmd_line = context.raw_data.get("cmd_line", {})
+    if "--explicit" not in sys.argv and not cmd_line.get("explicit"):
+        return
+    if "--no-pip" in sys.argv or not cmd_line.get("pip"):
+        return
```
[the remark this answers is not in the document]
jaimergp: I've seen this kind of suggestion before but never understood why. Isn't the `return` above enough to guarantee the lack of execution? To me, using `if` explicitly emphasizes the early return, but I can also see how using `elif` better reveals the conditional relationship between the statements.

conda_pypi/cli/list.py
```diff
@@ -0,0 +1,22 @@
...
+
+from .. import __version__
+from ..main import pypi_lines_for_explicit_lockfile
+
+
+def post_command(command: str):
+    if command != "list":
+        return
+    cmd_line = context.raw_data.get("cmd_line", {})
+    if "--explicit" not in sys.argv and not cmd_line.get("explicit"):
```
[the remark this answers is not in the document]
jaimergp: I wish we had a good API for the parsed args here. I want to use `context.raw_data` explicitly but I fear this is a data structure API we are not watching closely. See https://github.com/conda/conda/issues/13924.

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+    pd.load()
+    lines = []
+    python_record = list(pd.query("python"))
+    assert len(python_record) == 1
+    python_record = python_record[0]
+    python_details = {"version": ".".join(python_record.version.split(".")[:3])}
+    if "pypy" in python_record.build:
+        python_details["implementation"] = "pp"
+    else:
+        python_details["implementation"] = "cp"
```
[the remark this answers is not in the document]
jaimergp: I'm not aware of any other impls in conda-forge, or actually any system to query the implementation robustly across channels without executing code. I'll need to look further into this.

We do have `python_abi` and so on, but that's conda-forge specific, isn't it? Probably worth moving to its own utility function. I'll refactor this function a bit to accommodate the other changes too.

conda_pypi/cli/install.py
```diff
@@ -0,0 +1,111 @@
...
+                if not dist_info.is_dir():
+                    continue
+                name, version = dist_info.stem.split("-")
+                expected_hash = pkgs.get((name, version), {}).get("hash")
+                if expected_hash:
+                    algo, expected_hash = expected_hash.split(":")
+                    if (dist_info / "RECORD").is_file():
+                        found_hash = compute_record_sum(dist_info / "RECORD", algo)
+                        if expected_hash != found_hash:
+                            log.warning(
```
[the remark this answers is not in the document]
jaimergp: I think conda has a couple options to disable checks. I think we can honor those here. And actually should. Let me add those.

conda_pypi/main.py
```diff
@@ -168,3 +194,138 @@ def ensure_target_env_has_externally_managed(command: str):
...
+            if path.name == "WHEEL" and path.parent.suffix == ".dist-info":
+                for line in path.read_text().splitlines():
+                    line = line.strip()
+                    if ":" not in line:
+                        continue
+                    key, value = line.split(":", 1)
+                    if key == "Tag":
+                        wheel.setdefault(key, []).append(value.strip())
+                    else:
+                        wheel[key] = value.strip()
```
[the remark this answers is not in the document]
jaimergp: I dug in the docs and it uses RFC email headers, so `email.parser.HeaderParser` from stdlib can be used. It requires a bit of post-processing, quite similar to what we see in [`conda.common.pkg_formats.python`](https://github.com/conda/conda/blob/a261ade3d0a276a214bff314e54c9a9dbb8c282e/conda/common/pkg_formats/python.py#L561). That module would better be here in this project now. That whole thing is mostly there to support `conda list` listing `pypi` dependencies after all. We can refactor it out and finally remove the `pip_interop_enabled` thingy.

conda_pypi/main.py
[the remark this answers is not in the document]
jaimergp: I extended this check so it explicitly checks for `cpython` too, and falls back to None if no matches were found.

conda_pypi/main.py
jaimergp: True. Using `packaging.tags.parse_tag` now. Thanks!

conda_pypi/main.py
jaimergp: Done.

> Also the *.dir directory created by the --target option above needs cleaning up

It's never actually created, so it fails to clean up. I didn't add it.

conda_pypi/cli/install.py
```diff
@@ -0,0 +1,111 @@
...
+    if "--file" not in argv:
+        return []
+    pypi_lines = []
+    pypi_prefix = "# pypi: "
+    pypi_prefix_len = len(pypi_prefix)
+    for i, arg in enumerate(argv):
+        if arg == "--file":
+            pypi_lines += [
+                line[pypi_prefix_len:]
+                for line in Path(argv[i + 1]).read_text().splitlines()
```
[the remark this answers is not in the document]
jaimergp: I added a safety check just in case now.

conda_pypi/cli/install.py
```diff
@@ -0,0 +1,111 @@
...
+
+log = getLogger(f"conda.{__name__}")
+
+
+def _prepare_pypi_transaction(pypi_lines) -> dict[str, dict[str, str]]:
+    pkgs = {}
+    for args in pypi_lines:
+        args = shlex.split(args)
+        record_hash = None
+        if "--" in args:
```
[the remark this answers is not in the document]
jaimergp: I changed this part and provided an argparser instance.

conda_pypi/cli/install.py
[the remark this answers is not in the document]
jaimergp: Done.

conda_pypi/cli/list.py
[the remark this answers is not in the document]
jaimergp: Changed so we only use `context.raw_data` now.

conda_pypi/main.py
jaimergp: Changed to conditional extensions here in general.

jaimergp:
> I do want the CEP. For the original format too.

There you go: https://github.com/conda/ceps/pull/79

jaimergp:
Hey @jezdez, I think I addressed most of your comments now. The PR is getting a bit too big so we can address further changes in new PRs. WDYT? After this PR I feel comfortable moving things to `conda-incubator`.

jezdez (approved):

jezdez:
@jaimergp thank you for doing the rounds on the standardization, this'll be a net benefit!

Issue #58, opened 2025-08-18
Rename conda pupa to conda pypi


danyeaw:
This is part of an effort to breakdown the work in https://github.com/conda-incubator/conda-pypi/pull/54 so that it can be reviewed.

#### What is the idea?
Rename the existing conda pupa command to conda pypi to establish the primary user interface. This gives users access to the robust conda-pupa conversion functionality under the new conda-pypi branding.
#### Why is this needed?
The conda-pupa functionality has been successfully merged and is working. Users need access to this functionality under the conda pypi name to match the project branding.
#### What should happen?
Update conda_pypi/plugin.py to register conda pypi instead of conda pupa
Change the plugin registration from name="pupa" to name="pypi"
Update any help text or documentation references
Keep the existing Click-based CLI functionality as-is
#### Additional Context
This is a simple rename that immediately gives users access to the full conda-pupa functionality under the new name. The existing conda pupa command works well and just needs to be exposed as conda pypi.

PR #69, opened 2025-08-21
Merge conda_pupa folder with conda_pypi

### LINKED ISSUE #65: Merge the conda_pupa and conda_pypi folders
This is a follow up as the next step after #64 is merged, but before we work on #57.

### What is the idea?
Merge the conda_pupa and conda_pypi folders into a single unified package structure while maintaining the existing conda pip and conda pupa subcommands as separate entry points.
### Why is this needed?
The project currently has two separate folder structures that resulted from merging two previously independent projects. Now that they're part of the same codebase, they should be organized as a single cohesive package to simplify maintenance and deployment.
### What should happen?
Consolidate both conda_pupa/ and conda_pypi/ directories into one package structure. Update the plugin registration in pyproject.toml to use a single entry point that handles both subcommands. Both conda pip and conda pupa commands should continue to work exactly as they do today for users.
The merged plugin should register both subcommands through a single conda plugin hookimpl, following the pattern already used in conda_pypi where multiple commands are handled within one plugin. All existing CLI interfaces, command-line arguments, and functionality must remain unchanged from the user's perspective.
### Additional Context
The current pyproject.toml registers both plugins separately under [project.entry-points.conda]. The conda_pypi plugin already demonstrates how to handle multiple commands within a single plugin through its conda_subcommands() and conda_post_commands() implementations. This same pattern can be extended to include the pupa subcommand.
All existing functionality should be preserved including CLI interfaces, plugin hooks for post-commands, and supporting modules for package conversion and dependency resolution. Testing should verify that both conda pip install and conda pupa commands work as expected after the merge.

danyeaw:
Closes #65.

Scope of this PR is to merge the two folders (conda_pypi and conda_pupa) following #64. This is a minimum change to only merge the folders while keeping tests passing, the next step will be to rename the `conda pupa` command with `conda pypi` with #58.


ryanskeith (approved):
These merges all check out to me.

danyeaw:
@jaimergp, could you please take a look at the externally managed tests? They seem to be failing consistently on CI with the Ubuntu runners.

PR #94, opened 2025-09-09
Reorganizing project files

travishathaway:
I noticed that the post_command files were nested under the `conda_pypi.cli` module which didn't make much sense. They are now under the `conda_pypi.post_command` module. The `conda_pypi.cli` module has also been reorganized so that each subcommand now has its own module.


ryanskeith (approved):
These look good to me.

conda_pypi/post_command/install.py
```diff
@@ -0,0 +1,124 @@
...
+                    elif on_error == "error":
+                        errors.append(msg[0] % msg[1:])
+                    else:
+                        log.debug(*msg)
+    if errors:
+        errors = "\n- ".join(errors)
+        raise CondaVerificationError(f"PyPI packages checksum verification failed:\n- {errors}")
+
+
+def post_command(command: str):
```
ryanskeith, line 78: I would love to chat about this post_command sometime.  I want to understand better how these post_commands work.  My current mental model tells me that this is hooking into the `conda install` command.  If that is so then how do we assure we are invoking the post command?  Would calling it programmatically also invoke it?

PR #110, opened 2025-09-22
Add `--index-url` to `conda pypi install`

jaimergp:
- Closes https://github.com/conda-incubator/conda-pypi/issues/70

Only for `conda pypi install`. ~We can't add it to `convert` yet because `index-urls` is not supported by `pypa/build` (see https://github.com/pypa/build/issues/410).~ `conda convert` only handles local sources so far, so there's no point in adding a new index there because it will never be fetched.

### DIFF
--- a/conda_pypi/cli/install.py
+++ b/conda_pypi/cli/install.py
@@ -75,7 +75,7 @@ def execute(args: Namespace) -> int:
     prefix_path = Path(context.target_prefix)
 
     if args.index_urls:
-        index_urls = tuple(dict.fromkeys([*args.index_urls]))
+        index_urls = tuple(dict.fromkeys(args.index_urls))
         finder = get_package_finder(prefix_path, index_urls)
     else:
         finder = None


conda_pypi/cli/install.py
```diff
@@ -66,7 +74,17 @@ def execute(args: Namespace) -> int:
     """
     prefix_path = Path(context.target_prefix)
 
-    converter = convert_tree.ConvertTree(prefix_path, override_channels=args.override_channels)
+    if args.index_urls:
+        index_urls = tuple(dict.fromkeys((*DEFAULT_INDEX_URLS, *args.index_urls)))
```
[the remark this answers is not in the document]
jaimergp: Oh, thanks!

dholth:
In pupa we don't allow pypa/build to install its own dependencies. We use pypa/build and friends to check the dependencies for the build system, and then for the actual build. If any are missing we call `ensure_requirements()` which is a `conda install` call. This intercepts a place where it is really easy to pip install a (build) requirement automatically even if you want conda dependencies for most things.

https://github.com/conda-incubator/conda-pypi/blob/main/conda_pypi/build.py#L106-L110, in the [`build_pypa`](https://github.com/conda-incubator/conda-pypi/blob/main/conda_pypi/build.py#L90) function which is used for leaf nodes or the package you are developing locally.

`convert_tree` on the other hand only works with pre-built wheels, calls [`build_conda`](https://github.com/conda-incubator/conda-pypi/blob/main/conda_pypi/build.py#L126) which is wheel to conda conversion.

So at no point do we ask build to find and install anything.

Also, pupa always operates in the equivalent of pip's `--no-build-isolation` mode instead of creating a temporary environment to build a package into a wheel.

tests/cli/test_install.py
```diff
@@ -28,3 +34,21 @@ def test_cli_plugin(monkeypatch):
     assert pypi_subcommand.summary == "Install PyPI packages as conda packages"
     assert pypi_subcommand.action is not None
     assert pypi_subcommand.configure_parser is not None
+
+
+def test_index_urls(tmp_env, conda_cli, pypi_local_index):
+    print("XXXXX")
```
soapy1, line 40: It  looks like maybe this is leftover from some testing?
```suggestion
```

soapy1 (approved):
Looks great

conda_pypi/cli/install.py
```diff
@@ -50,7 +51,14 @@ def configure_parser(parser: _SubParsersAction) -> None:
...
         action="store_true",
-        help="Do not search default or .condarc channels. Will search pypi.",
+        help="Do not search default or .condarc channels. Will search PyPI.",
+    )
+    install.add_argument(
+        "-i",
+        "--index-url",
+        dest="index_urls",
+        action="append",
+        help="Add a PyPI index URL (can be used multiple times).",
```
jaimergp, line 61: @danyeaw, just wanted to make sure you are alright with us _not_ adding `--index`, but `-i / --index-url`.

conda_pypi/cli/install.py
```diff
@@ -66,7 +74,17 @@ def execute(args: Namespace) -> int:
     """
     prefix_path = Path(context.target_prefix)
 
-    converter = convert_tree.ConvertTree(prefix_path, override_channels=args.override_channels)
+    if args.index_urls:
+        index_urls = tuple(dict.fromkeys([*args.index_urls]))
```
jaimergp, line 78:
```suggestion
        index_urls = tuple(dict.fromkeys(args.index_urls))
```

conda_pypi/cli/install.py
danyeaw: @jaimergp yup, I'm good with it.

PR #111, opened 2025-09-22
More accurate help message for 'convert'

jaimergp:
While looking into https://github.com/conda-incubator/conda-pypi/pull/110, I realized `conad pypi convert --help` has misleading info. It only supports local paths for now.

### DIFF
--- a/conda_pypi/cli/convert.py
+++ b/conda_pypi/cli/convert.py
@@ -3,6 +3,7 @@ from pathlib import Path
 
 from conda.auxlib.ish import dals
 from conda.base.context import context
+from conda.exceptions import ArgumentError
 
 from conda_pypi import build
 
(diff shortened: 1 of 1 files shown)


danyeaw:
Hi @jaimergp, I would think we would want the `convert` command to take multiple formats:

1. Package names: requests, numpy, packaging
2. Version specifications: requests>=2.25.0, numpy==1.21.*
3. URLs to wheels: https://files.pythonhosted.org/packages/...
4. Local files: ./dist/mypackage-1.0.tar.gz, /path/to/package.whl

If it only support 4 now, we could change the help text for now, but we should open an issue to expand the convert command.

jaimergp:
Correct, we need issues for those features, but for now the help message will drive you to an error message :D

PR #126, opened 2025-09-26
Autogenerate the CLI Reference Commands with Sphinx

### LINKED ISSUE #89: Auto-generate CLI Reference Documentation from Python Code
#### What is the idea?
Automatically generate the CLI reference documentation from the actual Python argument parser definitions instead of maintaining it manually. This would ensure the documentation always stays in sync with the actual CLI interface.

#### Why is this needed?
Currently, the CLI reference documentation was manually created in #62 in order to do docs first approach. However, when new options are added or existing ones are modified, the documentation can easily get out of sync with the actual code. This leads to confusion for users and extra maintenance burden for developers. Manual documentation is also prone to errors and inconsistencies in formatting or descriptions.

#### What should happen?
The documentation build process should automatically extract CLI information from the actual argument parser definitions in conda_pypi/cli.py:

```
extensions = ['sphinx.ext.autodoc', 'sphinxcontrib.autoprogram']

# In your .rst/.md file
.. autoprogram:: conda_pypi.cli:configure_parser()
   :prog: conda pypi
```

#### Additional Context
Since conda-pypi already uses argparse for its CLI, the parser definitions contain all the necessary information (help text, argument names, types, etc.) to generate comprehensive documentation.

danyeaw:
Closes #89.


ryanskeith (approved):
These look good to me.  

An interesting note though:
After building locally my browser wouldn't automatically open the index.html for the folders so I had to click into them.  I am hoping the website will behave better so we should spot check this after it is deployed.

PR #132, opened 2025-10-01
Suppress output coming from solve

soapy1:
fixes: https://github.com/conda-incubator/conda-pypi/issues/97


conda_pypi/cli/install.py
```diff
@@ -93,9 +93,22 @@ def execute(args: Namespace) -> int:
...
+    converted_packages = [
+        str(pkg.to_simple_match_spec())
+        for pkg in packages_to_install
+        if pkg.channel.canonical_name == channel_url
+    ]
+    if converted_packages:
+        converted_packages_dashed = "\n - ".join(converted_packages)
+        print(f"Converted packages\n - {converted_packages_dashed}\n")
+    else:
+        print("No packages converted\n")
```
ryanskeith, line 109: Would we ever hit this case with `conda-pypi`?  Shouldn't there at least be one conversion? Maybe this is catching the case where we have a stashed conversion?

ryanskeith (approved):
This is a nice addition to suppress solver errors and at the same time provide information to the user about what is happening.

conda_pypi/cli/install.py
soapy1: This is the case when the package is already installed. Maybe this message is confusing. I'm going to back this little message out.

PR #148, opened 2025-10-22
[CI] Drop python 3.9 and add python 3.11, 3.12, 3.13

soapy1:
Part 1 to address https://github.com/conda-incubator/conda-pypi/issues/137

This updates the:
* pyproject.toml to use python >= 3.10
* pixi configuration to run tests in python 3.10, 3.11, 3.12, 3.13 (drops configuration for running in 3.9)
* ci test workflow to run against  python 3.10, 3.11, 3.12, 3.13 (drops configuration for running in 3.9)

Part 2 will address updating the ci for building and pushing packages to the conda-canary channel. To address the [feedback](https://github.com/conda-incubator/conda-pypi/issues/137):
> We saw some feedback from users trying to install conda-pypi in their base environment with Python 3.12 and they had to downgrade a Python version.

Note, will need conda to adopt python 3.14 before rolling support out to the CI here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["39", "310", "311"]
+        python-version: ["310", "311", "312", "313"]
         include:
           - os: macos-15-intel
             python-version: "310"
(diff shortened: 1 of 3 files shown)


jezdez:
> Note, will need conda to adopt python 3.14 before rolling support out to the CI here.

Just for the record, that is dependent on the Anaconda build team and their build-out of the dependency chain, we may not want to block on that.

soapy1:
@jezdez yep, this is good to go

PR #149, opened 2025-10-22
Replace pip subprocess with installer library for wheel unpacking

### LINKED ISSUE #104: Replace use of pip for wheel unpacking with the PyPA installer library
**What is the idea?**
Currently the project is depending on pip to do .conda package building as part of the `build_conda()` function. It would be much better to use a library to do package installation for wheel conversion rather than calling pip itself. [Installer](https://installer.pypa.io) provides a low level library to unpack a wheel for this use case.

**Why is this needed?**
Reduce subprocess calls and a dependency on pip itself.

**What should happen?**
Add installer as a dependency, and replace logic to unpack wheels with this library.

**Additional Context**
Thanks to @dholth's comments https://github.com/conda-incubator/conda-pypi/pull/54#issuecomment-3276036486. There are other uses of pip for the PackageFinder and VCS editable packages which will need to be resolved separately.

ryanskeith:
Fixes #104.

# Changes
- Modified install_installer() function to use custom installation scheme targeting build directory instead of system paths
- Updated build_conda() to use install_installer() instead of install_pip()
- Simplified installation scheme to use only purelib (equivalent to pip --target)

## Performance Impact
These benchmarks were done on the functions themselves and not the whole process but from what I saw: 
~85.0% faster wheel unpacking on average (20x+ speedup)
Small packages: ~99% faster (215x speedup)
Large packages: 78-87% faster (4-8x speedup)

The installer library was already a dependency so minimal changes were required.

### DIFF
--- a/conda_pypi/installer.py
+++ b/conda_pypi/installer.py
@@ -28,9 +28,9 @@ def install_installer(python_executable: str, whl: Path, build_path: Path):
     # See sysconfig documentation for more details on scheme keys.
     # https://docs.python.org/3/library/sysconfig.html#installation-paths
     scheme = {
-        "purelib": str(site_packages), # Pure Python packages
-        "platlib": str(site_packages), # Platform-specific packages
-        "scripts": str(build_path / "bin"), # Console scripts
+        "purelib": str(site_packages),  # Pure Python packages
+        "platlib": str(site_packages),  # Platform-specific packages
+        "scripts": str(build_path / "bin"),  # Console scripts
     }
 
     destination = SchemeDictionaryDestination(


conda_pypi/installer.py
```diff
@@ -22,9 +20,20 @@
...
 def install_installer(python_executable: str, whl: Path, build_path: Path):
     # Handler for installation directories and writing into them.
+    # Create site-packages directory if it doesn't exist
+    site_packages = build_path / "site-packages"
+    site_packages.mkdir(parents=True, exist_ok=True)
+
+    # Use minimal scheme to mimic pip --target: purelib, platlib, and scripts
+    scheme = {
+        "purelib": str(site_packages),
+        "platlib": str(site_packages),
```
soapy1: What are these `purelib` and `platlib` parameters?

conda_pypi/installer.py
```diff
@@ -22,9 +20,20 @@
...
+        "scripts": str(build_path / "bin"),
+    }
+
     destination = SchemeDictionaryDestination(
-        sysconfig.get_paths(),
-        interpreter=sys.executable,
+        scheme,
+        interpreter=str(python_executable),
         script_kind="posix",
```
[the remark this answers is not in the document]
ryanskeith: That's good to here.  @soapy1 thanks for checking.

soapy1 (approved):
Looks great! I also tested this on Linux + Windows and seems to work well for installing packages and upgrading/downgrading them. 🦜

conda_pypi/installer.py
```diff
@@ -34,10 +45,12 @@ def install_installer(python_executable: str, whl: Path, build_path: Path):
             destination=destination,
             # Additional metadata that is generated by the installation tool.
             additional_metadata={
-                "INSTALLER": b"amazing-installer 0.1.0",
+                "INSTALLER": b"conda-pypi",
```
jezdez: @ryanskeith Can/should we add the version number of conda-pypi here?

PR #163, opened 2025-11-07
Add codspeed benchmarks

soapy1:
This expands the test suite to add benchmarks using codspeed. This benchmarking includes:
* installing pypi packages with conda pypi
* converting the whole tree of pypi packages
* building a conda package from a pypi package

To run these test use the command:
```
$ pixi run benchmark
```

fixes https://github.com/conda-incubator/conda-pypi/issues/78

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
       # auto format Python codes within docstrings
       - id: blacken-docs
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.5
+    rev: v0.14.6
     hooks:
       # lint & attempt to correct failures (e.g. pyupgrade)
       - id: ruff


pyproject.toml
```diff
@@ -91,7 +92,8 @@ sphinx-reredirects = "*"
 sphinx-sitemap = "*"
 
 [tool.pixi.feature.test.tasks]
-test = 'python -c "from conda_pypi.python_paths import ensure_externally_managed as e; e()" && python -mpytest'
+test = 'python -c "from conda_pypi.python_paths import ensure_externally_managed as e; e()" && python -mpytest -m "not benchmark"'
+benchmark-test = 'python -c "from conda_pypi.python_paths import ensure_externally_managed as e; e()" && python -mpytest --codspeed'
```
jezdez:
```suggestion
benchmark= 'python -c "from conda_pypi.python_paths import ensure_externally_managed as e; e()" && python -mpytest --codspeed'
```

tests/test_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+    benchmark,
+):
+    """Benchmark convert_tree. This test overrides channels so the whole
+    dependency tree is converted.
+    """
+
+    def setup():
+        repo_dir = tmp_path_factory.mktemp(f"{'-'.join(packages)}-pkg-repo")
+        prefix = str(tmp_path_factory.mktemp(f"{'-'.join(packages)}"))
+        conda_cli("create", "--yes", "--prefix", prefix, "python=3.11")
```
kenodegard: Usually the only reason we install Python is so conda can find site-packages. Would this test still work if we mock it instead? Something like:

https://github.com/conda-incubator/conda-pypi/blob/c9af08c9fd0418ca819cdcb6615ca4c83f752890/tests/test_plugins.py#L34-L38

soapy1: This is a good suggestion! When I tried it out I found that python is required in the target prefix in order for the package finder to work.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+from conda.testing.fixtures import CondaCLIFixture
+from conda.models.match_spec import MatchSpec
+from conda.common.path import get_python_short_path
+
+from conda_pypi.convert_tree import ConvertTree
+from conda_pypi.downloader import find_and_fetch, get_package_finder
+from conda_pypi.build import build_conda
+
+
+# @pytest.mark.benchmark
```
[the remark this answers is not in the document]
soapy1: After removing these tests, the benchmarks take about 1.5 hours to run. This is still pretty long, but much better.

PR #183, opened 2025-12-01
Trial fix for rattler-solver demo

ryanskeith:
Seeing if adding a cleanup of tokio runtime (I had a panic locally) will help tests pass.

### DIFF
--- a/conda_pypi/convert_tree.py
+++ b/conda_pypi/convert_tree.py
@@ -28,6 +28,7 @@ from unearth import PackageFinder
 
 from conda_pypi.build import build_conda
 from conda_pypi.downloader import find_and_fetch, get_package_finder
+from conda_pypi.exceptions import CondaPypiError
 from conda_pypi.index import update_index
 from conda_pypi.utils import SuppressOutput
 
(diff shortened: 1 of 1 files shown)


conda_pypi/convert_tree.py
```diff
@@ -223,17 +203,22 @@ def convert_tree(
...
+                # maybe need to add repodata_fn, command args
             )
 
-            with get_spinner(self._get_converting_spinner_message(channels)):
-                changes = self._convert_loop(
-                    max_attempts=max_attempts, solver=solver, tmp_path=tmp_path
-                )
-
-            return changes
+            try:
```
ryanskeith: I added this try statement as it seemed that there was panic error with Tokio.  I was hoping that by removing the solver that it would have to restart and subsequent solves could be performed.

conda_pypi/convert_tree.py
```diff
@@ -124,14 +94,24 @@ def _convert_loop(
...
+                missing_packages.update(set(parse_solver_error(e.message)))
 
             for package in sorted(missing_packages - fetched_packages):
-                find_and_fetch(self.finder, wheel_dir, package)
-                fetched_packages.add(package)
+                # HACK: until we get conda 25.11.x
+                if str(package)[-1] == "=":
+                    package = MatchSpec(package.name)
+
+                try:
```
ryanskeith: Also added this try except for when packages can't be found on PyPI.  I am not sure if this is the best thing to do.  I did this to specifically catch things like openssl which seemed to have been broken on this branch with `conda-rattler-solver`.

soapy1: hmmm, in general I think this is an ok thing to do to ensure that we don't fail on a package that doesn't exist on pypi. But, this does not really fix the root of the issue discussed in https://github.com/conda-incubator/conda-pypi/pull/176#issuecomment-3590148420. The solver thould know that a good version of openssl is installed in the environment and not produce an unsatisfiable error.

PR #206, opened 2026-01-19
Respect conda json output setting

soapy1:
conda-pypi should not output non-json output if the user has enabled json output. Outputting non-json output may break the ability for other programs to read the output from conda.

For example,

__previously__, conda pypi would output non json text to the terminal even if the user enables `CONDA_JSON=true` or the `--json` flag.

```
$ pixi run conda pypi --json install imagesize -p /tmp/test

Installing environment
{
  "success": true,
  "actions": {
. . .
}
```

__now__, conda pypi will not ouput non-json text to the terminal
```
$ pixi run conda pypi --json install imagesize  

{
  "success": true,
  "actions": {
. . .
}
```

### DIFF
--- a/conda_pypi/cli/install.py
+++ b/conda_pypi/cli/install.py
@@ -142,7 +142,7 @@ def execute(args: Namespace) -> int:
         for pkg in packages_to_install
         if pkg.channel.canonical_name == channel_url
     ]
-    
+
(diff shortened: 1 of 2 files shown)


soapy1:
pre-commit.ci autofix

PR #215, opened 2026-01-22
Add support for converting wheels to conda directly from the CLI

JeanChristopheMorinPerso:
I noticed that the `conda pypi convert` command was saying that it could convert wheels directly to conda. After trying and looking at the code, the command only allowed to pass a directory that would be used to run an actual build.

This PR adds support for passing a wheel, matching the description. I also slightly modified the description to remove the mention of URLs. I guess that URLs could potentially be supported, but right now they aren't (see https://github.com/conda-incubator/conda-pypi/issues/113).

This code was mostly LLM generated. I fully reviewed it and tweaked it a bit. Considering that the diff is small, that seems pretty reasonable.


danyeaw:
Thanks @JeanChristopheMorinPerso!

PR #217, opened 2026-01-23
Use CondaPackageExtractor plugin hook

### LINKED ISSUE #209: Use `CondaPackageExtractor` plugin hook to register `whl_package` extractor plugin
Depends on : https://github.com/conda/conda/pull/15441 
Once the `CondaPackageExtractor` plugin hook is added to `conda`, `conda-pypi` should use this plugin hook instead of `CondaPreCommand` to register a plugin that allows extraction of `.whl` packages.

ForgottenProgramme:
Closes #209.


conda_pypi/package_extractor/__init__.py
```diff
@@ -0,0 +1,3 @@
+from conda_pypi.package_extractor.extract_whl import extract_whl_as_conda_pkg
+
+__all__ = ["extract_whl_as_conda_pkg"]
```
kenodegard: I don't think we need this re-export? Since it's unused we should remove it

conda_pypi/package_extractor/extract_whl.py
kenodegard: In this file my IDE/linter complains about `installer.install` not being callable because we do not explicitly call `import installer.install`, let's add this to the imports to resolve the error

kenodegard: I was wrong, the type checker was complaining about a non-issue

jezdez (approved):

PR #222, opened 2026-01-27
experimental: adds a new workflow for QA for testing PyPI packages with `conda pypi install`

### LINKED ISSUE #213: New GitHub Action for testing `conda pypi install` for wheels
### What's the idea?

We'd like to find more issues like the following before `conda-pypi` is widely adopted:

- https://github.com/conda-incubator/conda-pypi/issues/150

We'd also like to gain the ability to arbitrarily test and package from PyPI across the three major operating systems we support (Windows, macOS and Linux).

### What should be done?

To help in our QA efforts, we propose adding a new GitHub action that does the following:

- [ ] GitHub Action should be a separate workflow and only run when it's manually triggered
- [ ] The GitHub Action should accept a list of packages available on PyPI
- [ ] For each package, attempt to install this package into a separate environment
- [ ] If installation succeeds, verify the package installed correctly by running a python interpreter and importing the package
- [ ] For any failures, print the exception message so it can be inspected later

The actual implementation should be done inside of `pytest` so that we utilize our existing testing infrastructure. But, we need a way to dynamically parameterize these tests so that we can pass in arbitrary package names. Ideally, it would be best if there was a way for us to pass these in on the command line in the GitHub Action runners.

The test that installs and tests these packages should not be run during the normal test suite (i.e. we only want to manually trigger these). To achieve this, we should define a new `pytest` marker.

### Additional context

The following shows how you can add new command line options to `pytest` that can then be passed in to tests:

In `conftest.py`:

```python
def pytest_addoption(parser):
    parser.addoption(
        "--values",
        action="append",
        help="Values to parametrize tests with",
    )


def pytest_generate_tests(metafunc):
    if "value" in metafunc.fixturenames:
        values = metafunc.config.getoption("--values")
        if values:
            metafunc.parametrize("value", values)
```

In your test:
```python
def test_dynamic(value):
    assert value
```

travishathaway:
### Description

fixes: #213

I'm still personally unsure if this is actually useful and makes sense to merge, but I figured this would be a good place to have that conversation.

To see some runs, you can check out my fork:

- https://github.com/travishathaway/conda-pypi/actions/workflows/qa-wheels.yml

### DIFF
--- a/.github/workflows/qa-wheels.yml
+++ b/.github/workflows/qa-wheels.yml
@@ -1,0 +1,79 @@
+name: QA - Wheel Testing
+
+on:
+  workflow_dispatch:
+    inputs:
+      packages:
+        description: 'PyPI packages to test (one per line)'
+        required: true
+        type: string
+      python-version:
+        description: 'Python version to test'
+        required: false
+        default: '311'
+        type: choice
+        options:
+          - '310'
+          - '311'
+          - '312'
+          - '313'
+
+jobs:
+  qa-test:
+    name: QA Test - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+
+    env:
+      PIXI_ENV_NAME: test-py${{ inputs.python-version }}
+
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          fetch-depth: 0
+
+      - uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7 # v0.9.3
+        with:
+          environments: ${{ env.PIXI_ENV_NAME }}
+
+      - name: Fix Ubuntu configuration
+        if: startswith(matrix.os, 'ubuntu-')
+        run: sudo rm /etc/pip.conf
+
+      - name: Setup project
+        run: |
+          pixi reinstall --environment ${{ env.PIXI_ENV_NAME }} conda-pypi
+          pixi run --environment ${{ env.PIXI_ENV_NAME }} dev
+          echo "channels: [conda-forge]" > .pixi/envs/${{ env.PIXI_ENV_NAME }}/.condarc
+
+      - name: Parse package list and run QA tests
+        shell: bash
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/qa-wheels.yml
```diff
@@ -0,0 +1,79 @@
...
+          pixi run --environment ${{ env.PIXI_ENV_NAME }} python -mpytest $pytest_args
+
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: qa-test-results-${{ matrix.os }}-py${{ inputs.python-version }}
+          path: |
+            ${{ runner.temp }}/**/*.log
+          retention-days: 7
```
kenodegard: Unnecessary

tests/conftest.py
```diff
@@ -13,6 +13,16 @@
 HERE = Path(__file__).parent
 
 
+def pytest_addoption(parser):
+    """Register --qa-packages command line option"""
+    parser.addoption(
+        "--qa-packages",
+        action="append",
```
kenodegard: change this to "extend" and we can skip all of the argument formatting in the workflow

.github/workflows/qa-wheels.yml
```diff
@@ -0,0 +1,79 @@
...
+
+          # Build pytest command with --qa-packages for each package
+          pytest_args="-m qa -v --basetemp=${{ runner.os == 'Windows' && 'D:\\temp' || runner.temp }} tests/test_qa_wheels.py"
+          for pkg in "${packages[@]}"; do
+            # Trim whitespace
+            pkg=$(echo "$pkg" | xargs)
+            if [ -n "$pkg" ]; then
+              pytest_args="$pytest_args --qa-packages $pkg"
+            fi
+          done
```
kenodegard: Flatten it so it's more readable:

```suggestion
        env:
          BASETEMP: ${{ runner.os == 'Windows' && 'D:\\temp' || runner.temp }}
        run: |
          # Convert package list to --qa-packages arguments
          qa_args=$(echo "${{ inputs.packages }}" | xargs | sed 's/[^ ][^ ]*/--qa-packages &/g')

          # Run pixi/pytest
          pixi run --environment ${{ env.PIXI_ENV_NAME }} python -mpytest \
            -m qa -v \
            --basetemp="${BASETEMP}" \
            $qa_args \
            tests/test_qa_wheels.py
```

PR #226, opened 2026-01-29
Fix benchmark task to only run benchmark-marked tests

jezdez:
## Summary

The benchmark task was running all tests with `--codspeed` instead of only running tests marked with `@pytest.mark.benchmark`. This caused the CI to hang as it ran the full test suite during benchmark runs.

## Changes

Updated the benchmark task in `pyproject.toml` to use `-m "benchmark"` marker filter:

```diff
-benchmark= '... && python -mpytest --codspeed'
+benchmark= '... && python -mpytest -m "benchmark" --codspeed'
```

This matches the exclusion pattern used in the regular test task (`-m "not benchmark"`).

## Test plan

- [ ] CI benchmark job completes without hanging
- [ ] Only benchmark-marked tests are run during codspeed benchmarks

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -11642,8 +11642,8 @@ packages:
   timestamp: 1751548225624
 - pypi: ./
   name: conda-pypi
-  version: 0.3.1.dev48+ge9edf29a1.d20260128
-  sha256: a17c7b1b8f7e39a458e59a77a4f637bf8e4191b7b7cabbca0c668fb218776b01
+  version: 0.3.1.dev46+gc524f7109.d20260129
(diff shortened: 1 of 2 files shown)


kenodegard:
I believe `--codspeed` auto-selects for the benchmark marker:

`python -mpytest --codspeed` selects 5/80 tests to run:
Screenshot 2026-01-28 at 22 01 42

`pyton -mpytest -m "benchmark" --codspeed` also selects 5/80 tests to run:
Screenshot 2026-01-28 at 22 02 21

so it's not that CodSpeed runs all of the tests, the tests it runs are just incredibly slow

PR #242, opened 2026-02-06
Support test injection during convert

### LINKED ISSUE #228: Support injection of conda tests during `conda pypi convert`
### Checklist

### What is the idea?

Currently the `test` directory of a conda package produced using `conda pypi convert` is empty. This means that, although `conda build -t` doesn't cause an error, it doesn't run anything either. The desired state is to be able to run useful tests. A good place to start would be import tests, to allow some functionality without the problem expanding into a recipe-parsing problem.

### Why is this needed?

To be able to run tests on converted conda packages without bespoke surrounding logic.

### What should happen?

The converted conda package runs some useful tests when `conda build -t` is applied to it. Some suggestions for interface:

`conda pypi convert --tests <some-folder>` (simply copies the test folder into the package; breaks if not correct format)
`conda pypi convert --test-files test.yaml run_test.sh run_test.bat` (uses same format as conda recipes; more complex but more robust)
`conda pypi convert --import-tests imports.txt` (starts with import tests, can be extended later, simple but restricted functionality)

### Additional Context

Needed for internal project.

danpetry:
## Summary

Adds `--test-dir` flag to `conda pypi convert` to inject test files into generated conda packages, enabling test execution with `conda build -t`.

Closes #228

## Changes

This PR implements **Option 1** (folder copy approach) from the issue discussion:

- Added `--test-dir` CLI argument to `conda pypi convert`
- Test directory is copied to `info/test/` in the conda package
- Works with both wheel and source package conversions
- Includes input validation for the test directory path
- Validates that test directory contains at least one `run_test.*` file

## Test Plan

- [x] Tests pass for wheel conversion with test injection
- [x] Tests pass for source conversion with test injection  
- [x] Input validation rejects non-existent directories
- [x] Input validation rejects directories without `run_test.*` files
- [x] All existing tests continue to pass

## AI Usage Declaration

Claude Code was used to review the initial implementation, identify issues (missing source build support, lack of validation, missing documentation), implement fixes and enhancements, write tests, and draft this PR description. All code was reviewed by the author during the process.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


danyeaw (commented):
Hey @danpetry, thanks so much for the contribution. This will be a nice feature for sure!

danpetry:
thanks - how do I do that? I've had a look but can't figure it out

danpetry:
The CLA workflow was added by a bot PR last week and doesn't seem to be working

tests/cli/test_convert.py
```diff
@@ -43,3 +44,121 @@ def test_convert_wheel(tmp_path):
...
+    main_subshell(*args)
+
+    files = list(out_dir.glob("*.conda"))
+    assert files, f"No .conda artifacts found in {out_dir}"
+
+    assert files[0].is_file()
+    assert os.path.getsize(files[0]) > 0
+    assert "demo-package" in files[0].name
+
+    args = ["build", "-t", str(files[0])]
```
danyeaw: Hi @danpetry, I'm concerned about test speed and adding conda-build as a test dependency. On my local machine, this PR adds about a minute of test time.

Since we are already depending on conda-package-streaming, why don't we unpack the .conda file and check that the tests exist for most of the tests?

If we want to keep one `conda build -t` test for integration, could we make it an optional dependency and add a test marker to skip it if conda-build isn't installed?

danyeaw (commented):
Hi @danpetry, thanks for your patience as we got the CLA and pre-commit ironed out. These changes look great, I just have a couple of comments with the major one being about test speed.

tests/cli/test_convert.py
jezdez: I concur; I don't think we need to prove the full conda-build case here, since we can inspect the generated wheel file here without adding conda-build, which is a circular dependency we should prevent adding.

jezdez (changes requested):

conda_pypi/build.py
```diff
@@ -172,6 +174,18 @@ def build_conda(
         # Rewrite RECORD for any changed files
         update_RECORD(record_path, site_packages, direct_url_path)
 
+    if test_dir:
```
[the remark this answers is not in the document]
danyeaw: Hi @danpetry, yup, that sounds great!

danpetry:
oh I see, a maintainer needs to approve the CI run. Would someone mind triggering that please?

jezdez (approved):
with comments

danyeaw (approved):
Thanks @danpetry!

PR #253, opened 2026-02-25
Allow to override built-in pypi-to-conda name mapping

### LINKED ISSUE #252: Allow user to override the built-in `grayskull_pypi_mapping`
### Checklist

### What is the idea?

Allow the user to override the built-in default `grayskull_pypi_mapping` with a cli argument like `--override-name-mapping`, providing a json file to use instead of the build-in one.

### Why is this needed?

Current built-in mapping contains a map from pypi to conda-forge name. Some package names differ between conda-forge and anaconda official channel.

### What should happen?

User should be able to override the built-in map with a CLI argument.

### Additional Context

I'm currently working on this feature.

tombenes:
### Description

This closes #252 
Adds a `--name-mapping` option to use a custom `.json` file instead of the built-in `grayskull_pypi_mapping.json`.

### Checklist - did you ...

~~- [ ] Add / update necessary tests?~~

### DIFF
--- a/conda_pypi/cli/convert.py
+++ b/conda_pypi/cli/convert.py
@@ -1,12 +1,14 @@
 from tempfile import TemporaryDirectory
 from argparse import Namespace, _SubParsersAction
 from pathlib import Path
+import json
 
 from conda.auxlib.ish import dals
 from conda.base.context import context
 from conda.exceptions import ArgumentError
 
 from conda_pypi import build, paths
+from conda_pypi.translate import validate_name_mapping_format
 
 
 def configure_parser(parser: _SubParsersAction) -> None:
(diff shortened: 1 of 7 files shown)


conda_pypi/cli/convert.py
```diff
@@ -82,6 +90,15 @@ def execute(args: Namespace) -> int:
     output_folder = Path(args.output_folder).expanduser()
     output_folder.mkdir(parents=True, exist_ok=True)
 
+    # Load name mapping if overriden
+    # Use built-in by default
+    pypi_to_conda_name_mapping = None
+    if args.override_name_mapping is not None:
+        if not Path(args.override_name_mapping).exists():
```
danyeaw: `args.override_name_mapping` is already a Path object here.

danyeaw (changes requested):
Hi @tombenes, thanks for the contribution!

Could you please add tests and docs for the change? I think it would also be good if we did some JSON format checking to ensure the name mapping in the right format.

dholth:
Why not call it `--name-mapping`

danyeaw (commented):
Hi @tombenes, thanks so much for the updates - things are looking really good!

danyeaw (approved):
Thanks @tombenes, great contribution!

PR #264, opened 2026-03-02
Fix analyze step stuck with no code changes

danyeaw:
### Description

I noticed that PRs #260 and #262 were stuck on the analyze step. We were filtering on paths to not rerun all the tests, but then the analyze step was getting stuck. This PR uses the solution from the conda repo to filter changes using the dorny/paths-filter action. This will allow the analyze run to see the other jobs passed as allowed-skips.

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-pypi/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,12 +6,6 @@ on:
     tags:
       - "*"
   pull_request:
-    paths:
-      - ".github/workflows/test.yml"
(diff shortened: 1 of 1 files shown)


jezdez (approved):
Oh, huh!

PR #266, opened 2026-03-02
Make `test-package-with-data` a flat-layout package to restore skipped failing test

agriyakhetarpal:
### Description

This package has a src-layout and the `src/` folder is gitignored, which is why the test was failing. I've changed it to a flat-layout package and added its files to the worktree with this PR. I tested this on my fork in https://github.com/agriyakhetarpal/conda-pypi/pull/1 as well, and CI is green.

The other option is to continue having it as an src-layout package; I would not mind adding more specific rules to the gitignore file to adjust this. Please LMK!

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-pypi/blob/main/news/TEMPLATE)) for the next release's release notes?~
     

danyeaw (approved):
Thanks @agriyakhetarpal!

PR #289, opened 2026-03-30
Remove redundant package record versions

danyeaw:
### Description

@sukanya-pai pointed out during a repodata review, that we don't need each package record to have a `record_version`. It is defined as a root value already and it is an optional field.

### Checklist - did you ...


ryanskeith (approved):
Looks good to me.

PR #293, opened 2026-04-04
Refactor dependency installation in build.py

dholth:
### Description

A bug caused pypa/build to look for build requirements in the base conda environment. This would cause e.g. an editable install that requires `hatchling` to fail if `hatchling` was installed in the base environment. conda-pypi thinks it is available, but when it runs the project's build in the target environment it is not available. If e.g. `hatchling` was not installed in the base environment, then conda-pypi appears to work and installs it in the target environment.

Update handling to detect whether the "editable" requirements (or others) exist in the target prefix, not in the prefix that conda itself runs in.

Fix #281 

### Checklist - did you ...

### DIFF
--- a/conda_pypi/build.py
+++ b/conda_pypi/build.py
@@ -5,7 +5,6 @@ Create wheels from pypa projects.
 """
 
 import csv
-import itertools
 import json
 import logging
 import os
@@ -86,10 +85,6 @@ def json_dumps(object):
     Consistent json formatting.
     """
     return json.dumps(object, indent=2, sort_keys=True)
-
-
-def flatten(iterable):
-    return [*itertools.chain(*iterable)]
 
 
 def build_pypa(
--- a/conda_pypi/build.py
+++ b/conda_pypi/build.py
@@ -112,7 +112,7 @@ def build_pypa(
         """
         for _retry in range(2):
             try:
-                missing = dependencies.check_dependencies(requirements, prefix=prefix)
+                dependencies.check_dependencies(requirements, prefix=prefix)
                 break
             except dependencies.MissingDependencyError as e:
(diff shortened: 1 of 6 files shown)


conda_pypi/dependencies/pypi.py
```diff
@@ -43,7 +43,13 @@ def check_dependencies(requirements: Iterable[str], prefix: Path):
...
             check=True,
         )
-        missing = json.loads(result.stdout)
+        missing_raw = json.loads(result.stdout)
+        missing = []
+        for requirement in missing_raw:
+            if isinstance(requirement, str):
+                missing.append(requirement)
+            elif isinstance(requirement, list) and requirement:
+                missing.append(requirement[-1])
```
dholth, line 52: Other parts of the list are transitive dependencies, last item a leaf dependency

conda_pypi/build.py
```diff
@@ -108,21 +106,27 @@ def build_pypa(
...
-            missing = dependencies.check_dependencies(build_system_requires, prefix=prefix)
-            break
-        except dependencies.MissingDependencyError as e:
-            dependencies.ensure_requirements(e.dependencies, prefix=prefix)
-
-    log.debug(f"Installing requirements for build system: {missing}")
-    # does flatten() work for a deeper dependency chain?
-    dependencies.ensure_requirements(flatten(missing), prefix=prefix)
-
-    requirements = builder.check_dependencies(distribution)
```
dholth, line 123: This line that uses pypa/build (imported from the conda base environment) to check dependencies was the main issue. It's cool that conda runs in a different environment than the build/install, but this mode is foreign to pypa/* tools.

danyeaw (approved):
Thanks @dholth, this looks great! I pushed a change to remove the `flatten()` function since it is no longer used.

PR #300, opened 2026-04-09
Add licenses to info/licenses

danyeaw:
### Description
Fixes #299 by parsing the `License-File` field and adding it to `info/licenses`.

### Checklist - did you ...

- [X] Add / update outdated documentation?


conda_pypi/license_files.py
```diff
@@ -0,0 +1,91 @@
...
+    Paths to try for one ``License-File`` value: ``.dist-info/<path>``, then (if
+    ``listed_path`` is a single segment) ``.dist-info/licenses/<name>``, then
+    ``site-packages/<path>`` via the parent of ``.dist-info``.
+    """
+    dist_info_dir = dist_info_dir.resolve()
+    candidates: list[Path] = [dist_info_dir / listed_path]
+    if len(listed_path.parts) == 1:
+        candidates.append(dist_info_dir / "licenses" / listed_path)
+    candidates.append(dist_info_dir.parent / listed_path)
+    return candidates
```
jezdez: Two things here:

1. The `if len(listed_path.parts) == 1:` guard means multi-segment `License-File` entries (e.g. `License-File: docs/NOTICE`) never check `dist-info/licenses/docs/NOTICE`, which is where PEP 639 puts it. The `licenses/` candidate needs to be checked unconditionally.

2. Drop `dist_info_dir.parent / listed_path`. That resolves to e.g. `site-packages/LICENSE` which could belong to any package. Well-formed wheels always have license files inside `.dist-info/` (legacy) or `.dist-info/licenses/` (PEP 639), so this fallback isn't needed.

Minor: `dist_info_dir.resolve()` is called here and again in `flatten_name`. Resolve once at the top of `copy_into_info_licenses` instead.

```suggestion
def _license_file_lookup_paths(dist_info_dir: Path, listed_path: Path) -> list[Path]:
    """
    Candidate paths for one ``License-File`` value.

    Checks ``.dist-info/<path>`` (pre-PEP 639 / legacy wheels), then
    ``.dist-info/licenses/<path>`` (PEP 639, Metadata-Version 2.4+).
    """
    return [
        dist_info_dir / listed_path,
        dist_info_dir / "licenses" / listed_path,
    ]
```

conda_pypi/license_files.py
```diff
@@ -0,0 +1,91 @@
...
+        entry = raw_line.strip()
+        if not entry:
+            continue
+        listed_path = Path(entry)
+        for candidate in _license_file_lookup_paths(dist_info_dir, listed_path):
+            path = candidate.resolve()
+            if path.is_file() and path not in seen:
+                seen.add(path)
+                resolved.append(path)
+                break
```
jezdez, line 85: With the parent fallback gone, add a warning when no candidate matches so people know a declared file is missing. The `for/else` pattern works well here, the `else` runs when the loop finishes without `break`:

```suggestion
    resolved: list[Path] = []
    seen: set[Path] = set()
    for raw_line in metadata.get_all("License-File") or []:
        entry = raw_line.strip()
        if not entry:
            continue
        listed_path = Path(entry)
        for candidate in _license_file_lookup_paths(dist_info_dir, listed_path):
            path = candidate.resolve()
            if path.is_file() and path not in seen:
                seen.add(path)
                resolved.append(path)
                break
        else:
            log.warning("License-File '%s' declared in metadata but not found in %s", entry, dist_info_dir)
```

Needs `import logging` and `log = logging.getLogger(__name__)` at the top.

conda_pypi/license_files.py
```diff
@@ -0,0 +1,91 @@
...
+
+    dest_dir = info_dir / "licenses"
+    dest_dir.mkdir(parents=True, exist_ok=True)
+
+    def flatten_name(src: Path) -> str:
+        try:
+            rel = src.resolve().relative_to(dist_info_dir.resolve())
+            return str(rel).replace("\\", "/").replace("/", "__")
+        except ValueError:
+            return src.name
```
jezdez: This turns `licenses/LICENSE` into `licenses__LICENSE` which is surprising. If the file came from `dist-info/licenses/`, just compute the relative path from that `licenses/` dir. Files found directly in `dist-info/` keep their basename. That way a PEP 639 wheel with `dist-info/licenses/LICENSE` produces `info/licenses/LICENSE` as expected.

```suggestion
    def dest_name(src: Path) -> str:
        licenses_dir = dist_info_dir / "licenses"
        try:
            return str(src.resolve().relative_to(licenses_dir.resolve()))
        except ValueError:
            return src.name
```

jezdez (commented):
Good approach overall. Separate module, proper `importlib.metadata` API, both code paths, integration tests with a real wheel.

A few things to sort out before merge:

1. Multi-segment `License-File` paths (e.g. `docs/NOTICE`) don't resolve correctly because the `licenses/` lookup is gated to single-segment only
2. The parent-directory fallback (`dist_info_dir.parent`) should go, it can match files from other packages
3. `flatten_name` turns `licenses/LICENSE` into `licenses__LICENSE`, should preserve the relative path instead
4. `license_file` in `about.json` is deprecated per CEP 34, consider dropping
5. Minor: redundant `resolve()` calls and an extra `PathDistribution` read in `whl.py`

See inline comments.

conda_pypi/license_files.py
```diff
@@ -0,0 +1,91 @@
...
+
+    rel_paths: list[str] = []
+    for src, name in zip(resolved, dest_names, strict=True):
+        dest = dest_dir / name
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copy2(src, dest)
+        rel_paths.append(f"info/licenses/{name}")
+
+    if about is not None:
+        about["license_file"] = rel_paths[0] if len(rel_paths) == 1 else rel_paths
```
jezdez: `license_file` is deprecated per [CEP 34](https://conda.org/learn/ceps/cep-0034/). The license files in `info/licenses/` are self-describing.

Either drop this write or add a comment noting the deprecation. If keeping it for backward compat, always use a `list` rather than switching between `str` and `list` depending on count.

conda_pypi/package_extractors/whl.py
```diff
@@ -136,6 +139,15 @@ def _create_conda_metadata(
...
         write_as_json_to_file(info_dir / "index.json", index_json_data)
 
+        dist_infos = sorted(self.sp_dir.glob("*.dist-info"))
+        if dist_infos:
+            copy_into_info_licenses(
+                dist_infos[0],
+                info_dir,
+                PathDistribution(dist_infos[0]).metadata,
+                about=None,
+            )
```
jezdez: Minor: this re-reads and re-parses METADATA from disk, but the `WheelFile` source already has it available. Threading the existing metadata through `finalize_installation` / `_create_conda_metadata` would avoid the extra read.

Not blocking, just a tidiness thing.

docs/features.md
```diff
@@ -65,9 +65,17 @@ mapping of PyPI dependencies to conda equivalents and provides cross-platform
...
 across different operating systems and architectures.
 
+The wheel’s SPDX-style `License-Expression` (or legacy `License` field) is
+copied into conda metadata (`license` in `info/index.json` and `about.json`).
+When the wheel lists files under PEP 639 `License-File`, those files are also
+copied into `info/licenses/` in the `.conda` package, and `about.json` records
+their paths in `license_file`. Resolution checks `.dist-info/<path>`,
+`.dist-info/licenses/<basename>` for single-segment entries (as used by many
+current wheels, e.g. PyPI `packaging`), then the parent of `.dist-info`.
+
```
jezdez, line 74: If the parent fallback is dropped, update this to match. Resolution is just `.dist-info/<path>` (legacy) and `.dist-info/licenses/<path>` (PEP 639). Also `license_file` in `about.json` is deprecated per CEP 34, so the mention of recording paths there could be dropped.

```suggestion
The wheel's SPDX-style `License-Expression` (or legacy `License` field) is
copied into conda metadata (`license` in `info/index.json` and `about.json`).
When the wheel lists files under PEP 639 `License-File`, those files are also
copied into `info/licenses/` in the `.conda` package. Resolution checks
`.dist-info/<path>` (pre-PEP 639 wheels) and `.dist-info/licenses/<path>`
(PEP 639, Metadata-Version 2.4+).
```

conda_pypi/license_files.py
danyeaw: Hey @jezdez, great catch, I removed support for license_file.

conda_pypi/package_extractors/whl.py
danyeaw: Nice one, I updated this to use the wheel metadata directly instead of `PathDistribution`

danyeaw:
Hi @jezdez, thanks for the thorough review! I committed your suggestions and cleaned up the other items.

danpetry:
Thank you both for the rapid efforts. What would be the chances of a release by early next week (mon sometime ideally)? Is there anything I can help with to make that easier?

jezdez (approved):

danyeaw:
@danpetry, yup, I'll try to do a release today if I can

PR #306, opened 2026-04-13
Cache pypi metadata to conda index cache

### LINKED ISSUE #276: Expose API to store PyPI JSON metadata into the conda-index cache
### Checklist

### What is the idea?

Expose a method in index.py that accepts raw PyPI JSON, converts it to a v3 wheel record using `translate.py`, and stores it in the conda-index cache via `cache.store()`.

### Why is this needed?

External tools like conda-pypi-test should only need to fetch PyPI metadata and hand it off. Centralizing conversion and storage in conda-pypi means consumers become pure fetchers with no conversion logic of their own.

### What should happen?

- Expose a method such as `store_pypi_metadata(cache, pypi_json)` that converts and stores a single package record
- Update `update_index()` in index.py to call `ChannelIndex.index()` with `update_cache=False` and `update_only=True` so repodata generation works without local files on disk

### Additional Context

_No response_

soapy1:
### Description

Adds a function for adding data originating from pypi.org to the conda-index cache. conda-index can then use the cache to generate the v3 conda index, including wheels by setting the additional options 
```
repodata_v3=True,
update_only=True
save_fs_state=False
cache_kwargs={"package_extensions": CONDA_PACKAGE_EXTENSIONS + (".whl",)},
```

See the changes to `tests/conda_local_channel/generate_noarch_wheel_repodata.py` for an example.

### Other changes

* updates conda-index dep to 0.11.0
* regenerates the wheel test channel
* drops mac-intel tests (support for mac-intel conda-canary builds are dropped in [conda 26.3.0](https://github.com/conda/conda/releases/tag/26.3.0) via https://github.com/conda/conda/pull/15689/changes/15bcf3a985ff2f3a0b51c8fe3921dd81857fbe29) 

Depends on conda-index changes: https://github.com/conda/conda-index/pull/265
Fixes https://github.com/conda/conda-pypi/issues/276

### Checklist - did you ...


conda_pypi/index.py
```diff
@@ -19,3 +25,48 @@ def update_index(path):
...
+    ```
+    def cache_repodata_entry(cache: BaseCondaIndexCache, name: str, version: str) -> dict[str, Any] | None:
+        pypi_endpoint = f"https://pypi.org/pypi/{name}/{version}/json"
+        pypi_data = requests.get(pypi_endpoint)
+        if pypi_data.json() is None:
+            raise Exception(f"unable to process {name} {version}")
+        store_pypi_metadata(cache, pypi_data.json())
+    ```
+    """
+    repodata_entry = pypi_to_repodata_noarch_whl_entry(pypi_json)
```
jezdez, line 48: `pypi_to_repodata_noarch_whl_entry` can return `None` when there's no `-none-any.whl` (sdist-only, platform wheels), which would `TypeError` on the next line. Could we bail early with `return None` and let callers decide how to handle that?

conda_pypi/index.py
```diff
@@ -19,3 +25,48 @@ def update_index(path):
...
+                "size": repodata_entry["size"],
+                "mtime": repodata_entry.get("timestamp", 1),
+            }
+        ]
+    )
+
+    # must contain sha256 and md5 keys but values may be None
+    assert "sha256" in repodata_entry
+    if "md5" not in repodata_entry:
+        repodata_entry["md5"] = None
```
jezdez: Small thing: `assert` gets stripped under `python -O`, so raising explicitly feels safer here. Might also be worth checking the value isn't empty, since `pypi_to_repodata_noarch_whl_entry` defaults `sha256` to `""`.

```suggestion
    if not repodata_entry.get("sha256"):
        raise ValueError(
            f"PyPI payload for {repodata_entry.get('name')!r} is missing a sha256 digest"
        )
    repodata_entry.setdefault("md5", None)
```

conda_pypi/index.py
```diff
@@ -19,3 +25,48 @@ def update_index(path):
...
+
+    cache.store(
+        fn=cache.database_path(path),
```
jezdez: I think `SqliteCache.store()` already calls `self.database_path(fn)` internally, so this ends up applying the prefix twice. Works today because the default prefix is `""`, but could bite us on any backend that sets one.

```suggestion
        fn=path,
```

conda_pypi/index.py
```diff
@@ -19,3 +25,48 @@ def update_index(path):
...
+        size=repodata_entry["size"],
+        mtime=repodata_entry.get("timestamp", 1),
```
jezdez: Looks like this always resolves to `1` since `timestamp` is still commented out in `pypi_to_repodata_noarch_whl_entry`. Could we wire `upload_time` through there, or at least default to `0` so it's clearer no real mtime is tracked yet?

tests/test_index.py
```diff
@@ -0,0 +1,29 @@
...
+
+    pypi_endpoint = "https://pypi.org/pypi/fastapi/0.116.1/json"
+    pypi_data = requests.get(pypi_endpoint)
+    store_pypi_metadata(cache, pypi_data.json())
+
+    # packages from database
+    packages = cache.indexed_packages()
+
+    assert len(packages.packages_whl) == 1
+    assert "fastapi" in [pkg.get("name") for pkg in packages.packages_whl.values()]
```
jezdez: Would be nice to keep this offline, a saved copy of `https://pypi.org/pypi/fastapi/0.116.1/json` at e.g. `tests/data/pypi_json/fastapi-0.116.1.json` would make the test deterministic and work in air-gapped CI.

```suggestion
import json
from pathlib import Path

from conda_pypi.index import store_pypi_metadata
from conda_index.index import ChannelIndex
from conda_index.utils import CONDA_PACKAGE_EXTENSIONS


HERE = Path(__file__).parent
PYPI_JSON_FIXTURES = HERE / "data" / "pypi_json"


def test_store_pypi_metadata(tmp_path: Path):
    channel_index = ChannelIndex(
        tmp_path,
        "haswheels",
        repodata_v3=True,
        update_only=True,
        save_fs_state=False,
        write_current_repodata=False,
        cache_kwargs={"package_extensions": CONDA_PACKAGE_EXTENSIONS + (".whl",)},
    )
    cache = channel_index.cache_for_subdir("noarch")

    pypi_data = json.loads(
        (PYPI_JSON_FIXTURES / "fastapi-0.116.1.json").read_text()
    )
    store_pypi_metadata(cache, pypi_data)

    packages = cache.indexed_packages()
    assert len(packages.packages_whl) == 1
    assert "fastapi" in [pkg.get("name") for pkg in packages.packages_whl.values()]
```

Once `store_pypi_metadata` handles `None`, mind adding a second test with an sdist-only payload to lock that in?

tests/conda_local_channel/generate_noarch_wheel_repodata.py
```diff
@@ -5,65 +5,50 @@
...
+        write_current_repodata=False,
+        repodata_v3=True,
+        update_only=True,
+        save_fs_state=False,
+        cache_kwargs={"package_extensions": CONDA_PACKAGE_EXTENSIONS + (".whl",)},
+    )
+    cache = channel_index.cache_for_subdir("noarch")
+
+    for pkg_tuple in repodata_packages:
+        cache_repodata_entry(cache, pkg_tuple[0], pkg_tuple[1])
```
jezdez, line 64: Once `store_pypi_metadata` can return `None`, a quick log line when we skip would help, otherwise fixture regenerations would silently drop packages.

.github/workflows/test.yml
```diff
@@ -55,8 +55,6 @@ jobs:
         os: [ubuntu-latest, windows-latest]
         pixi-env: ["py310", "py311", "py312", "py313", "py314", "canary"]
         include:
-          - os: macos-15-intel
-            pixi-env: "py310"
           - os: macos-latest
```
jezdez, line 58: Does this relate to the `conda-index` bump, or is it a separate decision? Either way, a line in the PR description (or splitting it into its own PR) would help reviewers.

tests/test_editable.py
```diff
@@ -130,7 +130,7 @@ def test_build_in_env(tmp_path):
     )
 
     installed = [
-        record.name for record in PrefixData(prefix, pip_interop_enabled=True).iter_records()
+        record.name for record in PrefixData(prefix, interoperability=True).iter_records()
```
jezdez, line 133: Disregard my earlier question here, this one is a forced update: conda deprecated the `pip_interop_enabled` kwarg in 25.9 (`PrefixData.__init__` uses `@deprecated.argument("25.9", "26.3", "pip_interop_enabled", rename="interoperability")`), so callers have to move to `interoperability=True` before 26.3. Sorry for the noise.

news/306-add-pypi-metadata-to-conda-index-cache.md
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
```
jezdez: Small one, the template asks to drop unused sections otherwise the `<news item>` placeholders end up in the changelog. Could also be a good spot to mention the `conda-index` bump.

```suggestion
### Enhancements

* Add function to store pypi metadata in the conda-index cache. This will allow to seed the conda-index cache with pypi packages to include in repodata. (#276 via #306)
* Bump `conda-index` to `>=0.11.0` and regenerate the wheel test channel using `ChannelIndex.index(...)` with `repodata_v3=True`.
```

jezdez (changes requested):
Nice cleanup! Delegating to `ChannelIndex.index()` is a big improvement over the old hand-rolled writer, thanks for working through the `conda-index` bump.

A few things I'd like to sort out before this merges:

- CI is failing across the matrix on `tests/test_conda_local_channel.py::test_local_channel_repodata_no_per_package_record_version`, which expects `repodata_version == 3`. `conda-index` 0.11.0 keeps `1` at the top of v3 monolithic output and nests the v3 payload under `"v3"`, so the assertion would need to shift to something like `assert "v3" in repodata and "whl" in repodata["v3"]` while keeping the per-record `record_version` check.
- `store_pypi_metadata` trips on PyPI packages without a `-none-any.whl`, and also ends up double-applying `database_path` to the cache key. Details inline.
- The new test reaches out to `pypi.org`, would be nice to switch to a fixture (suggestion inline).
- Dropping `macos-15-intel` still looks unrelated, happy to be wrong there, but a note in the description would help reviewers.

Edit: `pip_interop_enabled` -> `interoperability` is a forced rename from conda 25.9's deprecation, not a drive-by. Thanks to @dholth for catching that.

dholth:
Are you sure? I've never seen rever include a <news item> ?

.github/workflows/test.yml
soapy1: Added a note in the pr description - support for mac-intel conda-canary builds are dropped in [conda 26.3.0](https://github.com/conda/conda/releases/tag/26.3.0) via https://github.com/conda/conda/pull/15689/changes/15bcf3a985ff2f3a0b51c8fe3921dd81857fbe29.

conda_pypi/index.py
soapy1: Good idea!

conda_pypi/index.py
soapy1: True. Looks like this is also consistent with the postgres backend.

conda_pypi/index.py
soapy1: Lets default this to 0 for now. I've created https://github.com/conda/conda-pypi/issues/328 to look into a better solution for this.

tests/test_index.py
dholth, line 1: test_index has the tests that originated in conda-build, some not ideal.

soapy1: hmmm, I not quite sure what you mean by this. Which test in particular are you concerned about? 

Each of the tests in test_index.py are meant to test the `conda_pypi.index.store_pypi_metadata` function in order to verify:
* packages get added to the set of indexed packages in the cache
* check for error conditions when the provided metadata can not be transformed into a repodata entry
* check for error conditions when the provided metadata is missing required information

jezdez (approved):

tests/test_index.py
dholth: I was thinking about the conda-index repository but this is conda-pypi.

PR #307, opened 2026-04-15
Don't subclass `installer.destinations.SchemeDictionaryDestination` any longer and set a minimum bound on installer v1

### LINKED ISSUE #286: Updates for `installer` v1 release
PyPA/installer v1 is finally out! The list of changes is at https://github.com/pypa/installer/releases/tag/1.0.0. In #246, I made some changes can now be rearranged, and possibly some miscellaneous improvements as well.

However, it is to be noted that we aren't currently pinning `installer` for our dependencies.

I'd be happy to be assigned to this and work on it. Thanks!

agriyakhetarpal:
### Description

Closes #286. This PR sets a lower pin for `installer` v1 and removes the workaround we had to introduce in #246 for path traversal in scenarios that arise with wheels containing duplicate headers (or any files in the directory schemes, for that matter). This has been fixed upstream in the latest release.

cc: @danyeaw

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-pypi/blob/main/news/TEMPLATE)) for the next release's release notes?~ These are internal changes with no behaviour change for the public API, and hence do not make sense for a NEWS file, but happy to add one if insisted!
     
### DIFF
--- a/conda_pypi/installer.py
+++ b/conda_pypi/installer.py
@@ -13,34 +13,9 @@ from conda.cli.main import main_subshell
 from conda.core.package_cache_data import PackageCacheData
 from installer import install
 from installer.destinations import SchemeDictionaryDestination
-from installer.records import Hash, RecordEntry
 from installer.sources import WheelFile
 
-from conda_pypi.utils import hash_as_base64url
-
 log = logging.getLogger(__name__)
-
-
-# We've seen some wheels placing identical files in multiple .data/ schemes
-# (e.g., pybind11-global duplicates headers in both data/include/ and
-# headers/). SchemeDictionaryDestination raises FileExistsError on the
-# second copy. Here, we record the already-written file instead.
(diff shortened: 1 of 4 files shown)


danyeaw (approved):
Hi @agriyakhetarpal, thanks for cleaning this up! These changes look great, thanks!

agriyakhetarpal:
Hi @danyeaw, as we discussed: I took a brief second look, and there is a minor behaviour difference indeed – the new code preserves the last copy because we overwrite instead of skipping. However, as I was mentioning, it is the responsibility of the build backend/distribution builder to ensure that duplicate files in installation directories are identical in content and that the hash in the RECORD and that of the file on disk match. It would otherwise be a malformed wheel; so I can confirm we're all good to merge here! Thanks!

PR #312, opened 2026-04-16
[WIP] Add support for PEP 794 (Metadata-Version: 2.5)

agriyakhetarpal:
### Description

This PR adds support for [PEP 794](https://peps.python.org/pep-0794/), which adds `Import-Name` and `Import-Namespace` fields to core metadata. This is only a draft for now, see pending items below:

- [x] Add Import-Name and Import-Namespace reading and checking for conflicts
- [x] Add Import-Name conflict detection for wheel packages, and the whl channel path
	- [x] Should work per transaction, and then also across history (for previous solves, so that existing installed packages in an environment are also used to validate this). i.e., this should work across operations.
- [x] Add import names to `about.json`
- [ ] Figure out how to deal with conda packages and PyPI packages with the same import name (kind of similar to point 2)
- [x] Add tests
- [x] Add docs

### Checklist - did you ...


dholth:
Neat!

PR #322, opened 2026-04-20
Fix nested license directory issue in conda packages

danpetry:
### Description

**What**

  - Use License-File metadata value directly as destination path instead of computing relative path from .dist-info/
  - Fixes double nesting when wheels use the
PEP 639 layout (.dist-info/licenses/): info/licenses/licenses/LICENSE → info/licenses/LICENSE

  **Why**

  When wheels use the PEP 639 layout (.dist-info/licenses/), the previous code computed the relative path from .dist-info/ which included the licenses/ prefix, causing double nesting that (although compliant) doesn't match conda package norms.

  **How**

  Track the original License-File entry alongside the resolved source path, then use the entry value as the destination path. This correctly handles both flat layout (older wheels) and PEP 639 layout.

  **Testing**

  - 5/5 license file tests pass
  - Updated test expectations to match correct behavior

  **AI Disclosure**

  AI used: Investigation of conda-forge package conventions (20+ packages), implementation, test updates, PR description
  Human involvement: Identified the issue, guided research direction, approved design approach and commit

### Checklist - did you ...


news/fix-license-directory-nesting.md
```diff
@@ -0,0 +1,3 @@
+### Bug fixes
```
danyeaw: Hey @danpetry, we should also fix this from the last PR, since I noticed we missed this in the review as well.

The news file should be a copy of the TEMPLATE. The filename should start with the PR number and should have no file extension. The unused sections of the news shouldn't be removed.

The end of the blurb should have the PR number as well, like `use the PEP 639 layout. (#322)`

Here is an example from conda: https://github.com/conda/conda/blob/main/news/15867-deepcopy

tests/test_license_files.py
```diff
@@ -52,10 +52,8 @@ def test_license_file_basename_in_dist_info_licenses(tmp_path: Path):
     meta = PathDistribution(dist_info_dir).metadata
     rel_paths = copy_into_info_licenses(dist_info_dir, info_dir, meta)
 
-    assert rel_paths == ["info/licenses/licenses/LICENSE"]
-    assert "Apache-2.0" in (info_dir / "licenses" / "licenses" / "LICENSE").read_text(
```
jezdez, line 56: 😬

conda_pypi/license_files.py
```diff
@@ -97,12 +97,11 @@ def copy_into_info_licenses(
     dest_dir.mkdir(parents=True, exist_ok=True)
 
     rel_paths: list[str] = []
-    for src in resolved:
-        rel = src.relative_to(dist_resolved)
-        dest = dest_dir / rel
+    for src, listed_path in resolved:
+        dest = dest_dir / listed_path
```
jezdez, line 103: Small hardening ask: we're taking the destination straight from metadata now, so a wheel with `License-File: ../../../foo/LICENSE` plus a matching file at `.dist-info/licenses/../../../foo/LICENSE` would copy outside of `info_dir/licenses/`. The old `relative_to(dist_resolved)` path would have raised on `..`, so this is a small regression in hardening.

```suggestion
    for src, listed_path in resolved:
        if listed_path.is_absolute() or ".." in listed_path.parts:
            log.warning("Skipping License-File %r with unsafe path segments", listed_path)
            continue
        dest = dest_dir / listed_path
```

Maybe add a test that asserts the skip while you're in there?

jezdez (changes requested):
Thanks for catching this @danpetry, good fix.

The double nesting came in via #300's follow-up commit 712290c "Fix multi-segment licenses, and drop dist_info.parent fallback". That commit replaced a `flatten_name` helper with `rel = src.relative_to(dist_resolved); dest = dest_dir / rel`, which did fix the `docs/NOTICE` case and the flattened filenames, but for PEP 639 wheels `src` resolves to `.dist-info/licenses/LICENSE`, so `rel` encoded the discovery location rather than the declared `License-File` value and we ended up at `info/licenses/licenses/LICENSE`.

Using the `License-File` entry itself as the destination (what this PR does) is the right shape: where the file was found during discovery shouldn't leak into where it lands.

One small ask before merge, see inline.

Also fwiw the tests that went in with #300 were asserting the buggy `info/licenses/licenses/LICENSE` output, which is why neither of us caught it on review. Worth a brief note in the test once you're in there.

dholth:
https://github.com/conda/conda-pypi/pull/324 could affect this approach. Instead of micromanaging a temporary directory to conda, override installer's destination `.write_to_fs()` to create a `.conda` without a temporary directory.  It doesn't directly solve the nested license directory problem yet.

conda_pypi/license_files.py
danpetry: Hmm. For our use case, we need the `License-File` in there. Even if we check afterwards, we'd prefer not to drop such cases. Can `dest` default to `dest_dir` in that case, and fail hard on collisions? Say `log.error` or throw an exception.

news/fix-license-directory-nesting.md
dholth: FYI rever will accept partial copies of TEMPLATE. It will also put `### My made up section (some news)` in its changelog, if any file in the news/ folder has sections not in TEMPLATE. I'm not sure if it cares about the filenames, but the perfect news filename is `number-short-description` i.e. `322-license-file-nesting`

danpetry: Fixed - renamed to `322-fix-license-directory-nesting` with full TEMPLATE format and PR number suffix.

conda_pypi/license_files.py
danpetry: Good catch on the path traversal risk. I've added validation that raises `ValueError` immediately when parsing metadata if the `License-File` contains absolute paths or `..` segments. Since legitimate wheels won't have these paths, failing loudly is the right behavior. Added tests for both cases.

danpetry:
This should be fine to review separately from https://github.com/conda/conda-pypi/pull/324. That PR's branch tests previous license-copying function still works; this changes the logic within it

danpetry:
@jezdez how's it looking here to you? Could I get a re-review or approval please?

tests/test_license_files.py
```diff
@@ -106,8 +104,32 @@ def test_license_file_multi_segment_under_licenses_subdir(tmp_path: Path):
...
-    assert (info_dir / "licenses" / "licenses" / "docs" / "NOTICE").read_text() == "Legal\n"
+    assert rel_paths == ["info/licenses/docs/NOTICE"]
+    assert (info_dir / "licenses" / "docs" / "NOTICE").read_text() == "Legal\n"
+
+
+@pytest.mark.parametrize(
+    "unsafe_path",
+    [
+        pytest.param("../LICENSE", id="parent-traversal"),
+        pytest.param("foo/../../../LICENSE", id="deep-traversal"),
```
jezdez, line 117: Nit: the docstring on `test_license_file_unsafe_path_raises` promises "path traversal or absolute paths", but the absolute-path branch of the guard (`listed_path.is_absolute()`) isn't exercised. Easy to cover with one more param:

```suggestion
        pytest.param("foo/../../../LICENSE", id="deep-traversal"),
        pytest.param("/etc/passwd", id="absolute-path"),
```

tests/test_license_files.py
```diff
@@ -52,10 +52,8 @@ def test_license_file_basename_in_dist_info_licenses(tmp_path: Path):
     meta = PathDistribution(dist_info_dir).metadata
     rel_paths = copy_into_info_licenses(dist_info_dir, info_dir, meta)
 
-    assert rel_paths == ["info/licenses/licenses/LICENSE"]
-    assert "Apache-2.0" in (info_dir / "licenses" / "licenses" / "LICENSE").read_text(
-        encoding="utf-8"
-    )
+    assert rel_paths == ["info/licenses/LICENSE"]
```
jezdez, line 57: Nit (and for `test_license_file_multi_segment_under_licenses_subdir` below too): worth a one-line comment here noting that the old `info/licenses/licenses/LICENSE` assertion from #300 was the buggy output this PR fixes, so future readers don't wonder why the expectation flipped. Something like:

```suggestion
    # #300 asserted the buggy info/licenses/licenses/LICENSE here; #322 fixes it.
    assert rel_paths == ["info/licenses/LICENSE"]
```

jezdez (approved):
Thanks for the iteration — the fix looks right and the path-traversal validation via `ValueError` is a clean resolution to the earlier thread. Approving.

Two tiny nits inline that you can fold in or skip; neither is blocking.

danyeaw (approved):

PR #319, opened 2026-04-17
Fix archive structure

danpetry:
### Description

NB this PR still needs manual checking and refinement.

## What                                                         
  - Fix CEP 35 violations: spurious `/` root entry and `info` directory misrouted to pkg archive                                                                                                                      
  - Handle symlinks and directories in info/ per CEP 34 path_type spec                                                                                                                                                
  - Add symlink target validation to prevent archive slip from untrusted wheels                                                                                                                                       
  - Add comprehensive CEP 34/35 compliance test                                                                                                                                                                       
                                                                                                                                                                                                                      
  ## Why                                                                                                                                                                                                              
  Generated `.conda` packages violated CEP 35 ("root level MUST match target location"), causing tar warnings and potentially triggering security scanners.                                                           
                                                                                                                                                                                                                      
  ## How                                                                                                                                                                                                              
  Replace `tar.add(build_path, "")` with iteration over children, handling info/ specially to route entries correctly to info_tar via CondaTarFile's `name.startswith("info/")` check. Symlinks are validated to
  ensure targets stay within build_path.                                                                                                                                                                              
                                                                                                                                                                                                                      
  ## Testing
  - New `test_conda_package_conforms_to_cep_34_35` validates all MUST requirements                                                                                                                                    
  - All existing tests pass                                                                                                                                                                                           
  - Linting clean                                                                                                                                                                                                     
                                                                                                                                                                                                                      
  ## AI Disclosure                                                                                                                                                                                                    
  **AI used:** Implementation, test writing, PR description                                                                                                                                                           
  **Human involvement:** Designed approach, reviewed all changes, approved architecture decisions, identified security concern (untrusted wheel symlinks) 

### Checklist - did you ...

<!-- If any of the following items aren't relevant to your contribution,
     please either tick them or use ~strikethrough~ so we know you've gone
     through the checklist. -->

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda-pypi/blob/main/news/TEMPLATE)) for the next release's release notes?
     <!-- All "significant" changes should get an entry:
            - user-facing changes or enhancements
            - bug fixes
            - deprecations
            - documentation updates
            - etc -->
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?


<!-- Just as a reminder, everyone in all conda org spaces (including PRs)
     must follow the Conda Org Code of Conduct (link below).

     Finally, once again, thanks for your time and effort. If you have any
     feedback in regards to your experience contributing here, please
     let us know!

[...]

PR #324, opened 2026-04-20
Create .conda from .whl with fewer temporary files

### LINKED ISSUE #317: Converted package contains unwanted entries in conda package
### Checklist

### What happened?

Hello, I tried converting a wheel into a conda package using `conda pypi convert` and I noticed two weird things:

1. The package contains an entry named `/`. This doesn't look right to me.
2. The `pkg-*` subarchive contains an empty `info/` folder.

```
$ unzip -p awesometkinter-2021.11.8-pypi_0.conda pkg-awesometkinter-2021.11.8-pypi_0.tar.zst | tar t -
/
info/
site-packages/
site-packages/awesometkinter/
site-packages/awesometkinter/__init__.py
site-packages/awesometkinter/__main__.py
site-packages/awesometkinter/bidirender.py
site-packages/awesometkinter/button.py
site-packages/awesometkinter/config.py
site-packages/awesometkinter/datepicker.py
site-packages/awesometkinter/dialog.py
site-packages/awesometkinter/frame.py
site-packages/awesometkinter/images.py
site-packages/awesometkinter/label.py
site-packages/awesometkinter/menu.py
site-packages/awesometkinter/progressbar.py
site-packages/awesometkinter/scrollbar.py
site-packages/awesometkinter/text.py
site-packages/awesometkinter/tooltip.py
site-packages/awesometkinter/utils.py
site-packages/awesometkinter/version.py
site-packages/awesometkinter-2021.11.8.dist-info/
site-packages/awesometkinter-2021.11.8.dist-info/INSTALLER
site-packages/awesometkinter-2021.11.8.dist-info/METADATA
site-packages/awesometkinter-2021.11.8.dist-info/RECORD
site-packages/awesometkinter-2021.11.8.dist-info/WHEEL
site-packages/awesometkinter-2021.11.8.dist-info/entry_points.txt
site-packages/awesometkinter-2021.11.8.dist-info/licenses/
site-packages/awesometkinter-2021.11.8.dist-info/licenses/LICENSE
site-packages/awesometkinter-2021.11.8.dist-info/top_level.txt
```

### Additional Context

As things are, a package that has `/` (or absolute paths) will not be installable with mamba. mamba explicitly disallows this via https://github.com/mamba-org/mamba/blob/4823aee40b38e5c17189a5f9b33e90b7971f74e1/libmamba/src/core/package_handling.cpp#L546

And in libarchive: https://github.com/libarchive/libarchive/blob/aca8b0d3d24e04f8ae8fe330014d22d754801fc4/libarchive/archive_write_disk_posix.c#L3234-L3242

This generates this error when trying to install a converted package with mamba:
```
critical libmamba Path is absolute
error    libmamba Error opening for reading "/Users/jcmorin/miniconda/pkgs/awesometkinter-2021.11.8-pypi_0/info/index.json": No such file or directory
error    libmamba Error when extracting package: [json.exception.parse_error.101] parse error at line 1, column 1: attempting to parse an empty input; check that your input string or stream contains the expected JSON
critical libmamba Found incorrect downloads. Aborting
```

dholth:
### Description

What if we didn't unpack the wheel just to put it into a conda?

This PR updates our `_CondaWheelDestination` to have a `write_to_fs()`, which the superclass uses for all filesystem writes. This should be good for performance, security, and controlling exactly what goes into our conda package.

Closes #317.

### Checklist - did you ...


conda_pypi/build.py
```diff
@@ -88,6 +96,48 @@ def json_dumps(object):
...
+    if not candidates:
+        message = f"No .dist-info directory found in wheel: {whl}"
+        raise FileNotFoundError(message)
+    if len(candidates) > 1:
+        message = f"Multiple .dist-info directories found in wheel: {whl} ({sorted(candidates)})"
+        raise ValueError(message)
+    return next(iter(candidates))
+
+
+def _extract_dist_info_dir(whl: Path, target_dir: Path, dist_info_name: str) -> Path:
```
dholth: should be merged with above function, open zipfile once

conda_pypi/build.py
```diff
@@ -143,50 +193,72 @@ def build_conda(
...
+                    continue
+                rel = path.relative_to(info_tmp).as_posix()
+                _add_tar_bytes(tar, f"info/{rel}", path.read_bytes())
+
+        # used especially for console_scripts
+        if link_json := metadata.link_json():
+            _add_tar_json(tar, "info/link.json", link_json)
+
+        # Allow pip to list us as editable or show the path to our project.
+        # XXX leaks path
```
dholth: perhaps we no longer leak path

conda_pypi/installer.py
```diff
@@ -2,19 +2,30 @@
...
 from installer.destinations import SchemeDictionaryDestination
-from installer.records import RecordEntry
+from installer.records import Hash, RecordEntry
 from installer.sources import WheelFile
+from installer.utils import Scheme, copyfileobj_with_hashing
+
+from conda_pypi.conda_build_utils import PathType
+
+if TYPE_CHECKING:
+    from tarfile import TarFile
```
dholth: uses TarFile interface; would also be capable of making a .tar.bz2 archive.

tests/test_build.py
```diff
@@ -112,6 +112,39 @@ def test_build_conda_copies_licenses_to_info_licenses(
...
+        sys.executable,
+        is_editable=False,
+    )
+
+    assert out_conda.is_file()
+
+    pkg_names = {member.name for _, member in package_streaming.stream_conda_component(out_conda)}
+    info_names = {member.name for _, member in package_streaming.stream_conda_info(out_conda)}
+
+    assert all(name.startswith("site-packages") for name in pkg_names)
```
dholth: not universally true but we are looking for any leakage of `/User/<me>/python...` e.g.

tests/test_build.py
```diff
@@ -112,6 +112,39 @@ def test_build_conda_copies_licenses_to_info_licenses(
     assert lic_payload == b"BSD-3-Clause placeholder license text\n"
 
 
+def test_build_conda_members_stay_in_info_component(
```
danpetry, line 115: We should add anything in https://github.com/conda/conda-pypi/pull/319/changes#diff-e7bc8c7d1837f65155df89f3b2662ca449b5800ee5d6b6b656c3564cb50d3e82R128 that this test doesn't cover

conda_pypi/installer.py
```diff
@@ -27,12 +38,61 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+        super().__init__(*args, **kwargs)
+        self.conda_builder = conda_builder
+        self.package_paths: list[dict] = []
+        self._members: set[str] = set()
+
     def write_script(self, name, module, attr, section):
         log.debug(f"Skipping script generation for {name} (handled via link.json)")
         return RecordEntry(path=name, hash_=None, size=None)
 
+    def write_to_fs(
```
danpetry, line 50: is this called/needed?

danpetry:
I like this approach for the speed, let's refine it to make sure the output is correct

conda_pypi/installer.py
dholth: The other `write_()` methods in https://github.com/pypa/installer/blob/main/src/installer/destinations.py#L196 delegate here.

conda_pypi/build.py
```diff
@@ -143,50 +196,76 @@ def build_conda(
...
+            package_paths = installer.install_installer_to_tar(python_executable, whl, tar)
+
+            # XXX set build string as hash of pypa metadata so that conda can re-install
+            # when project gains new entry-points, dependencies?
+
+            _add_tar_json(tar, "info/index.json", record)
+            _add_tar_json(tar, "info/about.json", metadata.about)
+
+            info_tmp = build_path / "info"
+            copy_into_info_licenses(dist_info, info_tmp, metadata.metadata)
```
dholth, line 200: may not behave with less-temporary-files strategy

conda_pypi/build.py
```diff
@@ -143,50 +196,76 @@ def build_conda(
...
+                        continue
+                    rel = path.relative_to(info_tmp).as_posix()
+                    _add_tar_bytes(tar, f"info/{rel}", path.read_bytes())
+
+            # used especially for console_scripts
+            if link_json := metadata.link_json():
+                _add_tar_json(tar, "info/link.json", link_json)
+
+            # Allow pip to list us as editable or show the path to our project.
+            # XXX leaks path
```
dholth: "leaks path" refers to e.g. the path to the checkout in the user's homedir, in editable mode. If project_path is absent in other cases then this is OK.

conda_pypi/installer.py
```diff
@@ -49,7 +110,7 @@ def install_installer(python_executable: str, whl: Path, build_path: Path):
         "headers": str(build_path / "include"),  # C/C++ headers (PEP 427 .data/headers/)
     }
 
-    destination = _CondaWheelDestination(
+    destination = SchemeDictionaryDestination(
```
dholth:
```suggestion
    destination = _CondaWheelDestination(
```

conda_pypi/installer.py
```diff
@@ -27,12 +38,62 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+                "sha256": hash_hex,
+                "size_in_bytes": size,
+            }
+        )
+
+        return RecordEntry(path, Hash(self.hash_algorithm, hash_), size)
+
 
-def install_installer(python_executable: str, whl: Path, build_path: Path):
+def install_installer(
```
dholth: Obsolete function? Update tests to use the  `_to_tar()` variant.

conda_pypi/build.py
```diff
@@ -88,6 +96,51 @@ def json_dumps(object):
     return json.dumps(object, indent=2, sort_keys=True)
 
 
+def _whl_dist_info_name(wheel_zip: zipfile.ZipFile) -> str:
+    candidates = {
+        name.split("/", 1)[0]
+        for name in wheel_zip.namelist()
```
dholth: This could be derived exactly from the wheel filename. See reference implementation.

conda_pypi/build.py
```diff
@@ -4,28 +4,37 @@
...
 import os
-import shutil
 import sys
+import tarfile
 import tempfile
+import zipfile
 from importlib.metadata import PathDistribution
 from pathlib import Path
-from typing import Union
+from typing import TYPE_CHECKING, Any, Union
```
dholth: these could probably go in the type checking block

conda_pypi/installer.py
```diff
@@ -27,33 +38,86 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+        is_executable: bool,
+    ) -> RecordEntry:
+        archive_path = os.path.join(self.scheme_dict[scheme], path)
+        archive_path = archive_path.replace(os.sep, "/")
+
+        if ".." in archive_path.split("/"):
+            raise ValueError(f"Path traversal detected: {archive_path}")
+
+        if archive_path in self._members:
+            if not self.overwrite_existing:
```
dholth: overwrite_existing should not be implemented in the write-to-archive case.

conda_pypi/installer.py
```diff
@@ -27,33 +38,86 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+        super().__init__(*args, **kwargs)
+        self.conda_builder = conda_builder
+        self.package_paths: list[dict] = []
+        self._members: set[str] = set()
+
     def write_script(self, name, module, attr, section):
         log.debug(f"Skipping script generation for {name} (handled via link.json)")
         return RecordEntry(path=name, hash_=None, size=None)
 
+    def finalize_installation(self, scheme, record_file_path, records):
```
dholth: RECORD should be written and INSTALLER should be added/say "conda"; check your ~/miniconda3/**/site-packages/

tests/test_installer_data_files.py
```diff
@@ -43,52 +44,43 @@ def test_package_wheel_path(tmp_path_factory):
...
-
-    with tmp_env("python=3.12") as prefix:
-        python_executable = Path(prefix, get_python_short_path())
-
-        installer.install_installer(
-            str(python_executable),
-            wheel_with_headers,
-            build_path,
+    """Wheel .data/headers/ files are recorded with include/ path."""
+    with tempfile.NamedTemporaryFile(suffix=".tar") as tar_file:
```
dholth: In tests we can use tmp_path/"filename" or an in-memory stream and no real tar file.

conda_pypi/build.py
```diff
@@ -88,6 +97,38 @@ def json_dumps(object):
...
+    info.mode = mode
+    tar.addfile(info, io.BytesIO(data))
+
+
+def _add_tar_text(tar: TarFile, name: str, text: str, mode: int = 0o664) -> None:
+    _add_tar_bytes(tar, name, text.encode("utf-8"), mode=mode)
+
+
+def _add_tar_json(tar: TarFile, name: str, payload: Any, mode: int = 0o664) -> None:
+    _add_tar_text(tar, name, json_dumps(payload), mode=mode)
```
jezdez: Please don't add these random functions, this is typical AI slop

danpetry:
Fix for windows failures here: https://github.com/conda/conda-pypi/pull/332
I can't push to this branch.

jezdez:
Two questions before I do a deeper pass:

First, on `_add_to_tar` — could this become a small wrapper class shared with `_CondaWheelDestination.write_to_fs`? Both are solving the same "write bytes into the conda tar" problem, but with subtly different rules: `write_to_fs` sets a file mode, rejects `..` segments, and dedupes members, while `_add_to_tar` does none of those, so the `info/` metadata files, licenses, tests, and `direct_url.json` all end up in the archive with mode `0` and no duplicate-member guard. Subclassing `TarFile` directly doesn't fit because `conda_builder` constructs the tar itself, but a composition wrapper would let both call sites converge on one code path. Related: is the `filter()` function at the top of `build.py` still pulling its weight now that nothing calls `tar.add(...)` on a directory anymore?

Second, on the "fewer temporary files" promise in the title — the license and test-dir blocks still write into `build_path` and read back, which seems to cut against the goal. Is a streaming variant of `copy_into_info_licenses` that yields name/bytes pairs in scope for this PR, or would you rather defer that to a follow-up?

dholth:
@jezdez I know there are many reasons not to convert an arch-specific package but consider outlier packages like the `tensorflow` wheel. 1.1G with 1.5M in `.dist-info`. This kind of package will benefit greatly from not unpacking the whole thing to a temporary directory.
We still get `SpooledTemporaryFile` for the two `pkg-` and `info-` inner archives because we need to know the final size before passing it to zstd, so that zstd includes its size header, so that clients know how much memory to allocate on decompression.
It would make sense for the license and test-dir blocks to copy from the zip to the conda directly but we could also defer that to another issue.

conda_pypi/installer.py
```diff
@@ -27,33 +38,87 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+        archive_path = os.path.join(self.scheme_dict[scheme], path)
+        archive_path = archive_path.replace(os.sep, "/")
+
+        if ".." in archive_path.split("/"):
+            raise ValueError(f"Path traversal detected: {archive_path}")
+
+        if archive_path in self._members:
+            message = f"File already exists: {archive_path}"
+            if self.overwrite_existing:
+                message = "{message} overwrite_existing not available in write-to-archive."
```
dholth:
```suggestion
                message = "{message}; overwrite_existing not available in write-to-archive."
```

conda_pypi/installer.py
```diff
@@ -27,33 +35,110 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+        path: str,
+        stream: BinaryIO,
+        is_executable: bool,
+    ) -> RecordEntry:
+        """
+        In installer==1.0.0, the SchemeDirectoryDestination() superclass
+        delegates all write_*() functions here.
+        """
+        archive_path = str(Path(self.scheme_dict[scheme], path).as_posix())
+        archive_path = archive_path.lstrip("/")
```
dholth:
```suggestion
```
possibly no longer necessary

conda_pypi/installer.py
```diff
@@ -27,33 +35,110 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+
+        if ".." in archive_path.split("/"):
+            raise ValueError(f"Path traversal detected: {archive_path}")
```
dholth, line 64: installer uses `Path.resolve()` and relative_to(base directory) but that checks the filesystem to follow symlinks etc.

conda_pypi/build.py
```diff
@@ -41,7 +46,7 @@ def filter(tarinfo):
 
 
 # see conda_build.build.build_info_files_json_v1
-def paths_json(base: Union[Path, str]):
+def paths_json(base: Path | str):
```
ryanskeith, line 49: Nice, we can go to this (instead of Union) now we are on Python 3.10.  Good call.

conda_pypi/installer.py
```diff
@@ -27,33 +35,110 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
     that breaks in other environments.
     """
 
+    conda_builder: tarfile.TarFile.TarFile
```
ryanskeith:
```suggestion
    conda_builder: tarfile.TarFile
```

conda_pypi/installer.py
```diff
@@ -27,33 +35,110 @@ class _CondaWheelDestination(SchemeDictionaryDestination):
...
+
+    def __init__(self, *args, conda_builder: tarfile.TarFile.TarFile, **kwargs):
```
ryanskeith:
```suggestion
    def __init__(self, *args, conda_builder: tarfile.TarFile, **kwargs):
```

ryanskeith (approved):
These look sound to me.  I am not seeing any show stoppers.  Please check the type hint suggestions.

danpetry:
@JeanChristopheMorinPerso checked this manually with `mamba` (which previously rejected converted conda packages) using the following commands

```
# Create local channel
$ mkdir -p  /tmp/asd/mychannel/noarch /tmp/asd/mychannel/osx-arm64
$ mv <conda package> /tmp/asd/mychannel/noarch/
$ cd /tmp/asd/mychannel
$ conda index --bz2 --zst .
# Now install the package with mamba (you can get mamba from conda-forge)
$ mamba create --prefix ./asd charset-normalizer -c /tmp/asd/mychannel -c https://repo.anaconda.com/pkgs/main

```

conda_pypi/build.py
dholth: The LLMs love outdated Python typing

conda_pypi/build.py
```diff
@@ -88,6 +93,12 @@ def json_dumps(object):
     return json.dumps(object, indent=2, sort_keys=True)
 
 
+def _add_to_tar(tar: tarfile.TarFile, name: str, data: bytes) -> None:
+    tar_info = tarfile.TarInfo(name)
+    tar_info.size = len(data)
+    tar.addfile(tar_info, io.BytesIO(data))
```
danyeaw:
```suggestion
def _add_to_tar(tar: tarfile.TarFile, name: str, data: bytes, mode: int = 0o644) -> None:
    # 0o644 matches conda-build's convention for info/ metadata files
    tar_info = tarfile.TarInfo(name)
    tar_info.size = len(data)
    tar_info.mode = mode
    tar.addfile(tar_info, io.BytesIO(data))
```

This was correct, but it wasn't obvious to me that the `info/` directory is 644, but the rest of the conda directory should be 664 in `write_to_fs()`. Adding a comment here and explicitly setting the mode would help explain this to my future self.

dholth: Sometimes directories have different permissions than files? But we aren't doing any explicit directories here? I suppose "it's conda-build's convention" is a reason.

tests/test_installer_data_files.py
```diff
@@ -43,51 +46,47 @@ def test_package_wheel_path(tmp_path_factory):
...
 def test_install_installer_headers(
     tmp_env: TmpEnvFixture,
     wheel_with_headers: Path,
-    tmp_path: Path,
 ):
-    """Wheel .data/headers/ files are installed to build_path/include/."""
-    build_path = tmp_path / "build"
-    build_path.mkdir()
+    """Wheel .data/headers/ files are added to include/ in tar members."""
+    tar = tarfile.TarFile("conda.tar", "w", fileobj=io.BytesIO())
```
dholth, line 76: A nice trick to make a tar but never write to disk.

PR #341, opened 2026-04-23
Add upload timestamp support

### LINKED ISSUE #340: store_pypi_metadata causes spurious re-extraction of virtual wheel packages
### Checklist

### What happened?

There is a mismatch in the mtime defaults used when writing virtual wheel package entries to the conda-index cache in `store_pypi_metadata`.

When `store_pypi_metadata` writes a package to the cache, it makes two separate writes, one for the filesystem state and one for the indexed state. Both use the timestamp field from the repodata entry to set the mtime, but since timestamp is not currently populated for PyPI packages, both fall through to their hardcoded defaults.

These defaults are different (1 vs 0), causing conda-index to treat every entry as "changed" on the next call to `channel_index.index()`. This triggers re-extraction of the virtual .whl files from disk, which don't exist, resulting in all packages being dropped from the generated repodata.json.

### Possible solutions:
Both writes in store_pypi_metadata should use the same mtime value so that conda-index's change detection finds no stale entries and skips re-extraction.

1. The simplest fix is to ensure the default mtime is consistent between the two calls.
2. Populate timestamp properly by converting PyPI's `upload_time` ISO string to a Unix integer.

### Additional Context

This issue was discovered when using store_pypi_metadata to generate a conda channel from PyPI wheel metadata at scale. The resulting repodata.json contained zero packages because every entry was being invalidated during indexing.

danyeaw:
### Description

Fixes #340 by adding upload timestamp support to package records.

I also did some tidying by refactoring our metadata conversion function in to a separate module and making `conda_pypi` imports consistently use absolute imports.

### Checklist - did you ...


conda_pypi/markers.py
```diff
@@ -163,80 +160,6 @@ def dependency_when(dependency: str, condition: str | None) -> str:
     return f"{dependency}[when={condition}]"
 
 
-def pypi_to_repodata_noarch_whl_entry(
```
danyeaw, line 166: I renamed this function and refactored it to the `pypi_metadata.py` module, no code changes.

tests/test_pypi_metadata.py
```diff
@@ -0,0 +1,148 @@
...
+        },
+    }
+    entry = pypi_to_repodata(pypi_data)
+    assert entry is not None
+    assert entry["name"] == "foo-bar"
+    assert entry["version"] == "1.0"
+    assert entry["subdir"] == "noarch"
+    assert entry["noarch"] == "python"
+    assert entry["fn"] == "foo_bar-1.0-py3-none-any.whl"
+    assert entry["timestamp"] == 1733499441000
```
danyeaw, line 74: This timestamp is new, but the overall test stayed the same.

tests/test_pypi_metadata.py
```diff
@@ -0,0 +1,148 @@
...
+    assert '[when="python<3.9"]' in te_dep
+    colorama_dep = next(d for d in entry["depends"] if d.startswith("colorama"))
+    assert '[when="__win"]' in colorama_dep
+
+    socks = entry["extra_depends"]["socks"]
+    assert len(socks) == 1
+    assert socks[0].startswith("pysocks")
+
+
+def test_pypi_to_repodata_timestamp_missing_upload_time():
```
danyeaw, line 87: This test is new

pyproject.toml
```diff
@@ -168,7 +168,13 @@ version-file = "conda_pypi/_version.py"
 line-length = 99
 
 [tool.ruff.lint]
-extend-select = ["I"] # sort imports
+extend-select = [
+  "I", # sort imports
+  "TID", # use absolute imports
```
dholth, line 173: Not sure about this lint, relative imports are handy if you want to vendor or rename a package, but I get the idea.

conda_pypi/index.py
```diff
@@ -57,7 +57,7 @@ def cache_repodata_entry(cache: BaseCondaIndexCache, name: str, version: str) ->
             {
                 "path": cache.database_path(path),
                 "size": repodata_entry["size"],
-                "mtime": repodata_entry.get("timestamp", 1),
+                "mtime": repodata_entry.get("timestamp", 0),
```
dholth, line 60: Occasionally you'll find some 'if timestamp' code which doesn't distinguish between None or 0. But this should be okay, I like this better than "1" if we don't send it to any functions that dislike 0.

conda_pypi/pypi_metadata.py
```diff
@@ -0,0 +1,101 @@
...
+
+from conda_pypi.markers import (
+    dependency_extras_suffix,
+    dependency_when,
+    extract_marker_condition_and_extras,
+)
+from conda_pypi.name_mapping import pypi_to_conda_name
+
+
+def pypi_to_repodata(
```
dholth, line 19: this must be similar to the METADATA to conda repodata function. Consolidate someday?

dholth (approved):
Thanks

pyproject.toml
danyeaw: We discussed this and you were ok with it.

PR #347, opened 2026-05-05
Add external-packages health check for conda doctor

### LINKED ISSUE #15582: Add `conda doctor --fix` for `external-packages`
### Checklist

### What is the idea?

Add an `external-packages` health check to `conda doctor` that detects packages installed via pip/uv and provides a `--fix` option to migrate them to conda-pypi.

### Why is this needed?

There are two main use cases for this:

- Pip packages can't be reliably locked in conda lockfiles.
- When you try to protect your base environment (see conda-incubator/conda-self#88), it clones base to a default environment. If base contains pip packages, the clone can no longer ensure your new environment matched what you had before.

 A health check would let them detect and fix the issue proactively.

### What should happen?

- Detection: Use `PrefixData.get_python_packages()` to find packages installed from PyPI. In verbose mode, list each package. User-facing copy should say "installed from PyPI" rather than "pip-installed".
- Fix: `conda doctor --fix` runs `conda install` using the standard channel configuration, pinning each package to its exact installed version. If a wheels channel is in the user's `channels` list, it participates in the solve like any other channel. The fix attempts a single solve for all packages at once. If any packages cannot be satisfied, they are identified and dropped, and the solve is retried for the remaining set. The user is shown which packages will be migrated and which will be skipped before any changes are made.
- `--fix` does not call PyPI directly. Packages that cannot be satisfied from the configured channels are skipped with a clear message.

Example UX (illustrative):

```text
$ conda doctor
✗ Found 3 packages installed from PyPI.
  Run 'conda doctor --fix' to migrate them using your configured channels.

$ conda doctor --verbose
✗ Found 3 packages installed from PyPI.
  Channels: my-wheels, conda-forge
  - requests 2.32.3
  - flask 3.1.0
  - custom-package 1.0.0

$ conda doctor --fix
Found 3 packages installed from PyPI. Recommend adding a wheel channel to conda: https://conda.github.io/conda-pypi/add-channel (this doesn't exist yet).
Migrate using channels [conda-pypi, conda-forge]? [y/N]: y
  ✓ requests
  ✓ flask
  ✗ custom-package: not available on configured channels

Summary: 2 migrated, 1 skipped
```

ForgottenProgramme:
### Description

- Introduces a `conda doctor` health check to detect Python packages installed with `pip` 
- Builds a migration plan that identifies pip-installed packages which have available equivalents in conda channels  using `pypi_to_conda_name` mapping
- Provides a safe migration flow that reinstalls eligible pip packages using conda, excluding unsafe/unavailable packages
- Ensures base environment protection by skiping migration in the base environment.
- Adds a post-migration cleanup step that removes stale .dist-info/.egg-info metadata not owened by conda

### Checklist - did you ...


jezdez (dismissed):
A few things to address here. The migration order is the main concern (see inline), plus some cleanup items.

Also the function is named `migrate_to_pypi` but it actually migrates *from* pip *to* conda. Rename to something like `migrate_to_conda` to avoid confusion.

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,116 @@
...
+
+from conda.env.pip_util import pip_subprocess
+from conda.common.constants import NULL
+from conda.base.constants import OK_MARK, X_MARK
+from conda.base.context import context
+from conda.cli.install import reinstall_packages
+from conda.core.prefix_data import PrefixData
+from conda.plugins import hookimpl
+from conda.plugins.types import CondaHealthCheck
+from conda.api import SubdirData
```
jezdez: `hookimpl`, `CondaHealthCheck`, and `Iterable` are imported but unused in this file. Remove them.

```suggestion
from conda.plugins import hookimpl
from conda.plugins.types import CondaHealthCheck
from conda.core.prefix_data import PrefixData
from conda.api import SubdirData
```

Actually, just delete lines 17-18 entirely since they're not used here.

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,116 @@
...
+
+if TYPE_CHECKING:
+    from argparse import Namespace
+    from collections.abc import Iterable
+
+    from conda.plugins.types import ConfirmCallback
+
+
+def find_external_packages(prefix: str) -> list[PrefixData]:
```
jezdez: Return type should be `list[PackageRecord]` or similar, not `list[PrefixData]`. `get_python_packages()` returns package records, not PrefixData instances.

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,116 @@
...
+
+    for pkg in packages:
+        name = pkg.name.replace("_", "-")
+
+        conda_name = pypi_to_conda_name(pkg.name)
+        if conda_name != name:
+            print(f"Note: '{name}' will be reinstalled as '{conda_name}' from conda channels.\n")
+    
+
+        # check if conda can install it
```
jezdez, line 62: `name` is computed with `replace("_", "-")` but then never actually used for the lookup, you pass `pkg.name` to `pypi_to_conda_name` not `name`. Also the print fires before you know whether the package is conda-installable, so it can be misleading. And there's trailing whitespace on line 61.

```suggestion
        conda_name = pypi_to_conda_name(pkg.name)
        if conda_name != pkg.name:
            print(f"Note: '{pkg.name}' will be reinstalled as '{conda_name}' from conda channels.")

        # check if conda can install it
        if conda_has_package(conda_name):
            safe.append(conda_name)
        else:
            external_only.append(name)
```

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,116 @@
...
+    for package in pypi_pkgs:
+        stdout, stderr = pip_subprocess(["uninstall", package, "-y"], prefix, cwd=None)
+
+        print(f"Uninstalling {package}...")
+        print(stdout)
+
+        if stderr:
+            print("Error:", stderr)
+
+    return reinstall_packages(args, safe_packages, force_reinstall=True)
```
jezdez: This is the main problem. You uninstall via pip first, then reinstall via conda. If the conda reinstall fails (network issue, solver conflict, package not actually in channels), the user loses the package entirely with no way back. Reverse the order: install with conda first, then uninstall the pip version.

```suggestion
    pypi_pkgs = [conda_to_pypi_name(pkg) for pkg in safe_packages]

    result = reinstall_packages(args, safe_packages, force_reinstall=True)
    if result != 0:
        return result

    for package in pypi_pkgs:
        stdout, stderr = pip_subprocess(["uninstall", package, "-y"], prefix, cwd=None)
        print(f"Uninstalling {package}...")
        if stdout:
            print(stdout)
        if stderr:
            print("Error:", stderr)

    return 0
```

tests/health_checks/test_external_packages.py
```diff
@@ -0,0 +1,37 @@
...
+
+    from conda.testing.fixtures import PipCLIFixture, TmpEnvFixture
+
+
+def test_no_external_packages(tmp_env: TmpEnvFixture):
+    with tmp_env() as prefix:
+        assert find_external_packages(prefix) == []
+
+
+py_ver = "3.10"
```
jezdez: Hardcoding `py_ver = "3.10"` will break on CI that doesn't have 3.10. Use the running interpreter version or a fixture.

```suggestion
py_ver = f"{sys.version_info.major}.{sys.version_info.minor}"
```

(with `import sys` at the top)

conda_pypi/health_checks/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Health checks for conda-pypi"""
```
jezdez: Missing trailing newline at end of file.

jezdez (dismissed):
A few things to address here. The migration order is the main concern (see inline), plus some cleanup items.

Also the function is named `migrate_to_pypi` but it actually migrates *from* pip *to* conda. Rename to something like `migrate_to_conda` to avoid confusion.

conda_pypi/health_checks/external_packages.py
ForgottenProgramme: I am now also storing the `pypi` names , so now it looks like this:

```
def build_migration_plan(packages):
    safe_pkgs_conda_names = []
    external_only = []
    safe_pkgs_pypi_names=[]

    for pkg in packages:
        name = pkg.name.replace("_", "-")

        conda_name = pypi_to_conda_name(pkg.name)
        if conda_name != name:
            print(f"Note: '{name}' will be reinstalled as '{conda_name}' from conda channels.\n")
    

        # check if conda can install it
        if conda_has_package(conda_name):
            safe_pkgs_conda_names.append(conda_name)
            safe_pkgs_pypi_names.append(name)
        else:
            external_only.append(name)

    return safe_pkgs_conda_names, external_only, safe_pkgs_pypi_names

```

ForgottenProgramme: @jezdez , if I am able to find the `conda` name for a `pypi` package, I think it is safe to assume that that package exists? 🤷🏼 
But even then it seems like a redundant check/ `if` statement: 
```
 if conda_name != name:
            print(f"Note: '{name}' will be reinstalled as '{conda_name}' from conda channels.\n")
 
```
I removed it. 👍🏼

conda_pypi/health_checks/external_packages.py
kenodegard: Wont calling `pip uninstall` remove many of the same files that conda installed?

I think it would be ideal if we don't do an uninstall at all. Instead forcefully overwrite the pip install with a conda install followed by extra handling/smarts to remove pip specific meta files?

ForgottenProgramme: @kenodegard, I tried to implement this approach again (skip uninstalling and simply reinstall with conda) and I can confirm that the behavior is not consistent. Sometimes `conda` uninstalls the pip package, sometimes it doesn't (even when a conda package of same or equivalent name is available), so that conda doctor is stuck in loop: it reports that there is a pip package, I run the `--fix`, it says it updated the package, but when you run conda doctor again, you again see the pip package being reported. I need to dig deeper if there is a discernable reason why it happens that way.

ForgottenProgramme: okay, I have found out what's happening. It's related to stale dist-info directories from previous `pip` installed versions of the packages and conda list picking those up instead of the latest installed package.

ForgottenProgramme: I have taken the approach to not use `pip uninstall` at all, and simply delete the stale dist-info directories after reinstall with conda.

ForgottenProgramme:
pre-commit.ci autofix

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,124 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Health check: Report packages not installed by conda"""
+
+from __future__ import annotations
+
+import shutil
+from pathlib import Path
```
jezdez: Needed by the ownership-aware cleanup below.

```suggestion
from pathlib import Path, PurePosixPath
```

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,124 @@
...
+    # there is only one site-packages directory in a conda env
+    site_packages_dir = site_packages_candidates[0]
+
+    # respect dist-info naming convention, replace dashes with underscores in package names
+    pkg_name = pkg_name.replace("-", "_")
+    pattern = f"{pkg_name}-{pkg_version}.dist-info"
+
+    for path in site_packages_dir.glob(pattern):
+        print(f"Removing stale file: {path}")
+        shutil.rmtree(path)
```
jezdez: Can we make the cleanup ownership-aware instead of matching by name/version? As written, same-version migrations can delete the dist-info directory conda just installed. This uses public `PrefixRecord.files` from the pre-fix pip record and the post-reinstall conda records, then only removes metadata dirs still not owned by conda.

```suggestion
def record_file_paths(record: PrefixRecord) -> tuple[PurePosixPath, ...]:
    """Return package file paths normalized to conda's manifest path style."""
    return tuple(PurePosixPath(path.replace("\\", "/")) for path in record.files)


def metadata_directories(record: PrefixRecord) -> set[PurePosixPath]:
    """Return Python metadata directories recorded for a package."""
    directories = set()
    for file_path in record_file_paths(record):
        for path in (file_path, *file_path.parents):
            if path.name.endswith((".dist-info", ".egg-info")):
                directories.add(path)
                break
    return directories


def clean_up_stale_files(prefix: str, package: PrefixRecord) -> None:
    """Remove pip metadata that remains unowned after conda migration."""
    prefix_path = Path(prefix)
    prefix_data = PrefixData(prefix, interoperability=False).reload()
    conda_owned_paths = {
        file_path
        for record in prefix_data.iter_records()
        for file_path in record_file_paths(record)
    }

    for metadata_dir in sorted(metadata_directories(package)):
        is_conda_owned = any(
            owned_path == metadata_dir or metadata_dir in owned_path.parents
            for owned_path in conda_owned_paths
        )
        if is_conda_owned:
            continue

        path = prefix_path.joinpath(*metadata_dir.parts)
        if path.is_dir():
            print(f"Removing stale metadata: {path}")
            shutil.rmtree(path)
```

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def migrate_to_conda(prefix: str, args: Namespace, confirm: ConfirmCallback) -> None:
```
jezdez: Fixer returns integer exit codes on all paths.

```suggestion
def migrate_to_conda(prefix: str, args: Namespace, confirm: ConfirmCallback) -> int:
```

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,124 @@
...
+
+    args.use_local = False
+    args.file = []
+    args.repodata_fns = ("repodata.json",)
+    args.update_modifier = NULL
+
+    reinstall_packages(args, safe_pkgs_conda_names, force_reinstall=True)
+
+    for pkg in safe_pkgs_pypi_names:
+        clean_up_stale_files(prefix, pkg.name, pkg.version)
```
jezdez: Don't remove metadata if the reinstall failed; otherwise a failed migration can hide the pip package from future detection without installing the conda equivalent.

```suggestion
    result = reinstall_packages(args, safe_pkgs_conda_names, force_reinstall=True)
    if result != 0:
        return result

    for pkg in safe_pkgs_pypi_names:
        clean_up_stale_files(prefix, pkg)

    return 0
```

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,158 @@
...
+            args,
+            safe_pkgs_conda_names,
+            force_reinstall=True,
+        )
+    except Exception as e:
+        print(f"Failed to reinstall packages with conda: {e}")
+        return 1
+
+    for pkg in safe_pkgs_pypi_names:
+        clean_up_stale_files(prefix, pkg)
```
jezdez: Good catch on `install()` returning `None` on success. Can we still normalize the result before cleanup? This treats `None` and `0` as success, but skips metadata cleanup if `reinstall_packages()` ever returns another exit code.

```suggestion
    try:
        result = reinstall_packages(
            args,
            safe_pkgs_conda_names,
            force_reinstall=True,
        )
    except Exception as e:
        print(f"Failed to reinstall packages with conda: {e}")
        return 1

    if result not in (None, 0):
        return result

    for pkg in safe_pkgs_pypi_names:
        clean_up_stale_files(prefix, pkg)
```

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,158 @@
...
+    return bool(result)
+
+
+def build_migration_plan(packages) -> list:
+    """Determine which packages can be safely migrated to conda."""
+    safe_pkgs_conda_names = []
+    safe_pkgs_pypi = []
+
+    for pkg in packages:
+        # name = pkg.name.replace("_", "-")
```
kenodegard:
```suggestion
```

conda_pypi/health_checks/external_packages.py
kenodegard: the type annotation `conda.cli.install.reinstall_packages -> int` is wrong

looking at the source code `conda.cli.install.reinstall_packages` _only_ returns `None` for success, all other error states raise an exception

conda_pypi/health_checks/external_packages.py
```diff
@@ -0,0 +1,158 @@
...
+
+    print("Cleaning up stale metadata directories...")
+    prefix_path = Path(prefix)
+    prefix_data = PrefixData(prefix, interoperability=False).reload()
+
+    conda_owned_paths = {
+        file_path
+        for record in prefix_data.iter_records()
+        for file_path in normalize_conda_file_paths(record)
+    }
```
kenodegard: this re-generates a set of **all** conda owned paths (for all installed packages) every time this cleanup is called but cleanup is called per package that was reinstalled, should we just get the list of conda owned paths for the current `prefix_record`?

conda_pypi/health_checks/external_packages.py
ForgottenProgramme: This seems a bit unneccesary 🤔

ForgottenProgramme:
@kenodegard 
>the type annotation conda.cli.install.reinstall_packages -> int is wrong
looking at the source code conda.cli.install.reinstall_packages only returns None for success, all other error states raise an exception

I also found the same when I looked at the install func. 
The "fixer" of a health check is defined to return an int. And the reinstall_packages func is returned by the fixers of several health checks. I guess that's why it was made to return an int

Should we fix that first? 

update:
opened an issue: https://github.com/conda/conda/issues/16170

conda_pypi/health_checks/external_packages.py
ForgottenProgramme: I will look at this now.

ForgottenProgramme:
pre-commit.ci autofix

ForgottenProgramme:
pre-commit.ci autofix

kenodegard (approved):

danyeaw:
Hey @ForgottenProgramme, it looks like this needs a news entry.

PR #352, opened 2026-05-06
Disable EXTERNALLY-MANAGED file during beta, show notice instead

### LINKED ISSUE #353: Replace EXTERNALLY-MANAGED hook with forward-looking warning
### What is the idea?

Replace the post-command hook that writes the PEP 668 EXTERNALLY-MANAGED marker file with a forward-looking warning notice for the duration of the community beta. The warning should not fire when the target environment is the base environment.

### Why is this needed?

Bundling conda-pypi with conda 26.5.0 gives the beta maximum community reach, but placing the EXTERNALLY-MANAGED file immediately blocks the very common conda activate && pip install pattern in user workflows and CI pipelines. A warning gives users advance notice before GA enforcement lands without breaking any existing workflow.

### What should happen?

- On conda create and conda install, when pip is present in a non-base target environment, conda-pypi logs a warning informing users that a future release will enforce PEP 668 pip protection and directing them to conda pypi install.
- No EXTERNALLY-MANAGED file is written to any environment during the beta period.
- The original `ensure_target_env_has_externally_managed` function is preserved in main.py for re-enablement after the beta.
- Skip during conda-build, build environments are ephemeral and should never show user-facing notices.

#### Context

conda-pypi places a PEP 668 `EXTERNALLY-MANAGED` marker file into each conda
environment via a post-command plugin hook (`ensure_target_env_has_externally_managed`
in `conda_pypi/main.py`). When present, this file causes pip to refuse `pip install` with an error, directing users to `conda pypi install` or `--break-system-packages` instead.

The community beta (targeting conda 26.5.0, launching May 12, 2026) bundles conda-pypi so that all conda users receive the plugin automatically. A GitHub search of `.github/workflows` files containing both `conda activate` and `pip install` confirms that mixing these two commands inside CI pipelines is extremely common. Blocking those workflows without warning will break existing setups. We also know that most users (do we have data on this?) use pip inside conda environments.

An initial discussion happened on Mar 6, 2026 between @dashagurova and @danyeaw, and resulted in the interim position that conda-pypi should not add the `EXTERNALLY-MANAGED` file to every environment for the beta. A linked conda issue was opened to pass `--break-system-packages` automatically when conda processes an `environment.yml` with a `pip:` section (see https://github.com/conda/conda/issues/15764). However, we now need a decision about this with an implementation to support the beta launch.

Four options are under consideration for the beta launch.

Options

##### Option 1: Remove conda-pypi from conda 26.5.0 — require manual opt-in

Users must run `conda install conda-pypi` to activate the plugin for the beta.

Tradeoffs

Benefit: Zero disruption to existing `pip install` workflows. The `EXTERNALLY-MANAGED` risk is limited to users who consciously chose the plugin. No CI pipelines break on upgrade.

Cost: Dramatically reduces beta reach. Users most likely to benefit from conda-pypi are the same users currently running `conda activate && pip install`. Users will not discover the plugin unless they seek it out. The goal of broad community validation before GA is undermined.

##### Option 2: Keep conda-pypi bundled, disable the EXTERNALLY-MANAGED file

conda-pypi ships with conda 26.5.0 but `ensure_externally_managed()` in `conda_pypi/python_paths.py` is disabled (the call removed from the post-command hook or guarded by a feature flag). The `EXTERNALLY-MANAGED` file is never written to user environments during the beta period.

Tradeoffs

Benefit: Full beta reach with no breakage to `pip install` workflows or CI pipelines. Gives users access to `conda install <package>` from a wheel channel without taking away their existing pip escape hatch. Straightforward to implement since it is a small, reversible change to `conda_pypi/main.py`.

Cost: The primary safety guarantee of conda-pypi is absent during beta. Users can continue mixing pip and conda without any warning. Re-enabling the file at GA or after the beta creates a future migration moment that will again risk breaking setups. The value proposition of `conda install <package>` over raw pip is harder to explain when pip still works freely.

##### Option 3: Patch pip (and uv pip) to downgrade the error to a warning

Rather than writing the `EXTERNALLY-MANAGED` file and fully blocking pip, intercept or monkey-patch pip's PEP 668 enforcement path so that it emits a warning and continues instead of raising an error.

Slightly longer term, this could also be something we try to add to pip, to allow warnings to be passed in instead of only errors. However, there is a risk that this won’t be accepted.

Tradeoffs

Benefit: conda-pypi can be bundled with full reach, the `EXTERNALLY-MANAGED` file remains in place (preserving the signal for future hard enforcement), and existing `pip install` commands continue to work while users see a visible warning nudging them toward `conda install <package>` with the new beta channel.

Cost: Patching third-party tools (pip, uv) is fragile. The patch must be re-validated on every `pip` release and would add short-term maintenance burden to the feedstocks, or conda would need to intercept or wrap calls to `pip` and `uv pip`. `uv` does not expose the same extension points as pip, making parity uncertain. This approach may also condition users to ignore the warning, potentially weakening the eventual hard error at GA. There is a risk that patching `pip` wouldn’t be accepted by the conda-forge feedstock maintainers.

##### Option 4: Keep conda-pypi bundled, replace the hook with a forward-looking warning

Proposed in #352 by @jezdez. The `ensure_target_env_has_externally_managed` post-command hook in `plugin.py` is replaced with a new hook, `notify_externally_managed_future`, that logs a `logger.warning` on `conda create` and `conda install` when pip is present in the environment. No `EXTERNALLY-MANAGED` file is written. The original function is preserved in `main.py` so it can be re-enabled after the beta. The warning text informs users that a future conda release will enforce PEP 668 protection and points them toward `conda install <package>` as the preferred alternative.

This is distinct from Option 2 (silent removal) in that users receive proactive notice during the beta rather than encountering the hard block for the first time at GA.

Tradeoffs

Benefit: Full beta reach with no breakage to `pip install` workflows or CI pipelines. Users receive advance notice and a call to action before enforcement lands, reducing the surprise factor at GA. The original enforcement hook is preserved in the codebase, making re-enablement a one-line change rather than a feature rebuild. Does not require patching any third-party tool.

Cost: Users may dismiss or ignore a warning that appears on every `conda create`/`conda install`, leading to warning fatigue before the hard error arrives. The safety gap is the same as Option 2 during the beta window, conda environments remain unprotected from accidental `pip install`. The warning text must be kept accurate and actionable across the beta period, any change to the migration path (e.g., `conda install` CLI) requires updating the warning.

#### Decision

Option 4 is accepted. conda-pypi will be bundled with conda 26.5.0. The `ensure_target_env_has_externally_managed` post-command hook will be replaced with `notify_externally_managed_future`, which logs a warning when pip is present
in a non-base target environment. The `EXTERNALLY-MANAGED` file will not be written during the beta period. The original enforcement function is preserved in `main.py` for re-enablement after beta.

#### Consequences

- No existing `pip install` workflow or CI pipeline breaks when conda-pypi ships with conda 26.5.0.
- Users receive advance notice during the beta that PEP 668 enforcement is coming, reducing the surprise factor when the hard block lands at GA.
- The warning fatigue risk is accepted as a worthwhile tradeoff given the beta timeline and the need for broad community reach.
- Re-enabling full enforcement after beta requires restoring the original hook registration in `plugin.py` (a one-line change) with the implementation already preserved in `main.py`.

jezdez:
### Description

Closes #353.

During the community beta (targeting conda 26.5.0), conda-pypi's `EXTERNALLY-MANAGED` marker file breaks existing `pip install` workflows and CI pipelines. This PR replaces the post-command hook that writes the PEP 668 marker file with a `logger.warning` notice that informs users about upcoming pip protection.

**What changes:**

- The `ensure_target_env_has_externally_managed` post-command hook is replaced in `plugin.py` with `notify_externally_managed_future`
- Instead of writing the `EXTERNALLY-MANAGED` file, a warning is logged on `conda create` and `conda install` (when pip is present): *"A future conda release will use PEP 668 to prevent 'pip install' in conda environments..."*
- The original `ensure_target_env_has_externally_managed` function is preserved in `main.py` for re-enablement after the beta period

**Why:**

- Bundling conda-pypi as a `run` dependency gives the beta maximum reach for community validation
- Placing the `EXTERNALLY-MANAGED` file would break the extremely common `conda activate && pip install` pattern in CI and user workflows
- The warning gives users advance notice and points them to conda-pypi's native alternatives (wheel extractor, `conda pypi install`)
- Full `EXTERNALLY-MANAGED` enforcement is planned for a later release once migration tooling is available (see conda/conda#13863, Phase 2)

**Related:**

- Epic: conda/conda#13863
- PEP 668 / EXTERNALLY-MANAGED discussion: conda/conda#15764
- ADR-001: EXTERNALLY-MANAGED File Strategy for conda-pypi Community Beta

### Checklist - did you ...

### DIFF
--- a/conda_pypi/main.py
+++ b/conda_pypi/main.py
@@ -189,6 +189,31 @@ def ensure_target_env_has_externally_managed(command: str):
         raise ValueError(f"command {command} not recognized.")
 
 
+def notify_externally_managed_future(command: str):
+    """
+    Beta-period post-command hook that prints a one-time notice about upcoming
+    EXTERNALLY-MANAGED enforcement instead of placing the marker file.
+    """
+    # Build environments are ephemeral; never show user-facing notices.
+    if os.environ.get("CONDA_BUILD_STATE") == "BUILD":
+        return
+    # Only notify in non-base environments where pip interop is relevant.
+    base_prefix = Path(context.conda_prefix)
+    target_prefix = Path(context.target_prefix)
+    if base_prefix == target_prefix or base_prefix.resolve() == target_prefix.resolve():
+        return
+    # No point warning about pip protection if pip isn't installed.
+    prefix_data = PrefixData(target_prefix)
+    if not list(prefix_data.query("pip")):
+        return
+    logger.warning(
+        "A future conda release will use PEP 668 to prevent 'pip install' "
+        "in conda environments. conda-pypi lets you install PyPI packages "
+        "natively with conda. "
+        "See: https://conda.github.io/conda-pypi/"
+    )
+
+
 def pypi_lines_for_explicit_lockfile(
     prefix: Path | str, checksums: Iterable[Literal["md5", "sha256"]] | None = None
 ) -> list[str]:
--- a/conda_pypi/main.py
+++ b/conda_pypi/main.py
@@ -191,7 +191,7 @@ def ensure_target_env_has_externally_managed(command: str):
 
 def notify_externally_managed_future(command: str):
     """
-    Beta-period post-command hook that prints a one-time notice about upcoming
+    Beta-period post-command hook that logs a warning about upcoming
     EXTERNALLY-MANAGED enforcement instead of placing the marker file.
     """
     # Build environments are ephemeral; never show user-facing notices.
@@ -207,10 +207,14 @@ def notify_externally_managed_future(command: str):
     if not list(prefix_data.query("pip")):
         return
     logger.warning(
-        "A future conda release will use PEP 668 to prevent 'pip install' "
-        "in conda environments. conda-pypi lets you install PyPI packages "
-        "natively with conda. "
-        "See: https://conda.github.io/conda-pypi/"
+        "\n"
+        "  This environment has pip installed. A future conda release will\n"
+        "  protect conda environments from accidental 'pip install' usage.\n"
+        "  Try the beta to install PyPI packages natively with conda:\n"
+        "    conda config --set solver rattler\n"
+        "    conda config --append channels conda-pypi\n"
+        "    conda install <package>\n"
+        "  More info: https://docs.conda.io/projects/conda/en/stable/new-features.html"
     )
 
 
--- a/conda_pypi/plugin.py
+++ b/conda_pypi/plugin.py
@@ -4,7 +4,7 @@ from conda.plugins import hookimpl
 from conda.plugins.types import CondaPackageExtractor, CondaPostCommand, CondaSubcommand
 
 from conda_pypi import cli, post_command
-from conda_pypi.main import ensure_target_env_has_externally_managed
+from conda_pypi.main import notify_externally_managed_future
 from conda_pypi.package_extractors.whl import extract_whl_as_conda_pkg
 
 
@@ -21,9 +21,9 @@ def conda_subcommands():
 @hookimpl
 def conda_post_commands():
     yield CondaPostCommand(
-        name="conda-pypi-ensure-target-env-has-externally-managed",
-        action=ensure_target_env_has_externally_managed,
-        run_for={"install", "create", "update", "remove"},
+        name="conda-pypi-notify-externally-managed-future",
+        action=notify_externally_managed_future,
+        run_for={"install", "create"},
     )
     yield CondaPostCommand(
         name="conda-pypi-post-install-create",
--- a/conda_pypi/plugin.py
+++ b/conda_pypi/plugin.py
@@ -23,7 +23,7 @@ def conda_post_commands():
     yield CondaPostCommand(
         name="conda-pypi-notify-externally-managed-future",
         action=notify_externally_managed_future,
-        run_for={"install", "create"},
+        run_for={"install", "create", "env_create"},
     )
     yield CondaPostCommand(
         name="conda-pypi-post-install-create",
--- a/news/disable-externally-managed-beta
+++ b/news/disable-externally-managed-beta
@@ -1,0 +1,7 @@
+### Enhancements
+
+* Disable `EXTERNALLY-MANAGED` file placement during the community beta period. Instead, log an informational notice about upcoming pip protection when environments with pip are created or updated. (#350)
+
+### Deprecations
+
+* The `EXTERNALLY-MANAGED` file will be re-enabled in a future release once migration tooling is available. (#350)
--- a/news/disable-externally-managed-beta
+++ b/news/disable-externally-managed-beta
@@ -1,7 +1,7 @@
 ### Enhancements
 
-* Disable `EXTERNALLY-MANAGED` file placement during the community beta period. Instead, log an informational notice about upcoming pip protection when environments with pip are created or updated. (#350)
+* Disable `EXTERNALLY-MANAGED` file placement during the community beta period. Instead, log an informational notice about upcoming pip protection when environments with pip are created or updated. (#353 via #352)
 
 ### Deprecations
 
-* The `EXTERNALLY-MANAGED` file will be re-enabled in a future release once migration tooling is available. (#350)
+* The `EXTERNALLY-MANAGED` file will be re-enabled in a future release once migration tooling is available. (#353 via #352)
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Includes:
 - `conda pypi install -e .`: Converts a path to an editable `.conda` format package.
 - `conda pypi convert`: Convert PyPI packages to `.conda` format without installing them.
 - `conda install` from wheel channels (experimental): channels can serve pure Python wheels directly in `repodata.json`.
-- Adds `EXTERNALLY-MANAGED` to your environments.
+- A warning when running `conda create` or `conda install` with `pip` in the environment.
 
 ## Why?
 
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -6,6 +6,7 @@ from conda.testing.fixtures import CondaCLIFixture, TmpEnvFixture
 from pytest_mock import MockerFixture
 
 from conda_pypi import plugin
+from conda_pypi.main import notify_externally_managed_future
 from conda_pypi.package_extractors import whl
 
 WHL_HTTP_URL = "https://files.pythonhosted.org/packages/45/7f/0e961cf3908bc4c1c3e027de2794f867c6c89fb4916fc7dba295a0e80a2d/boltons-25.0.0-py3-none-any.whl"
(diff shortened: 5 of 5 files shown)


dholth:
Thanks for promulgating the pending pip protection printout.

jezdez (commented):
Added some code comments explaining the warning logic.

conda_pypi/main.py
```diff
@@ -189,6 +189,31 @@ def ensure_target_env_has_externally_managed(command: str):
         raise ValueError(f"command {command} not recognized.")
 
 
+def notify_externally_managed_future(command: str):
+    """
+    Beta-period post-command hook that prints a one-time notice about upcoming
```
danyeaw:
```suggestion
    Beta-period post-command hook that logs a warning about upcoming
```

conda_pypi/main.py
```diff
@@ -189,6 +189,31 @@ def ensure_target_env_has_externally_managed(command: str):
...
+    # No point warning about pip protection if pip isn't installed.
+    prefix_data = PrefixData(target_prefix)
+    if not list(prefix_data.query("pip")):
+        return
+    logger.warning(
+        "A future conda release will use PEP 668 to prevent 'pip install' "
+        "in conda environments. conda-pypi lets you install PyPI packages "
+        "natively with conda. "
+        "See: https://conda.github.io/conda-pypi/"
+    )
```
danyeaw, line 218:
```suggestion
    logger.warning(
        "\n"
        "  This environment has pip installed. A future conda release will\n"
        "  protect conda environments from accidental 'pip install' usage.\n"
        "  Try the beta to install PyPI packages natively with conda:\n"
        "    conda config --set solver rattler\n"
        "    conda config --append channels conda-pypi\n"
        "    conda install <package>\n"
        "  More info: https://docs.conda.io/projects/conda/en/stable/new-features.html"
    )
```

Reformats from:

```
# To deactivate an active environment, use
#
#     $ conda deactivate
WARNING conda.conda_pypi.main:notify_externally_managed_future(209): A future conda release will use PEP 668 to prevent 'pip install' in conda environments. conda-pypi lets you install PyPI packages natively with conda. See: https://conda.github.io/conda-pypi/
```

To:
```
# To deactivate an active environment, use
#
#     $ conda deactivate

    conda install <package>
```

news/disable-externally-managed-beta
```diff
@@ -0,0 +1,7 @@
+### Enhancements
+
+* Disable `EXTERNALLY-MANAGED` file placement during the community beta period. Instead, log an informational notice about upcoming pip protection when environments with pip are created or updated. (#350)
+
+### Deprecations
+
+* The `EXTERNALLY-MANAGED` file will be re-enabled in a future release once migration tooling is available. (#350)
```
danyeaw:
```suggestion
* Disable `EXTERNALLY-MANAGED` file placement during the community beta period. Instead, log an informational notice about upcoming pip protection when environments with pip are created or updated. (#353 via #352)

### Deprecations

* The `EXTERNALLY-MANAGED` file will be re-enabled in a future release once migration tooling is available. (#353 via #352)
```

danyeaw (commented):
Hi @jezdez, this looks great and I did some manual testing to create an environment with it. I suggested some improvements to the logged warning based on my testing.

Should we add a couple of tests and make sure the docs / README are accurate about externally managed as well?

I'm also going to open a conda issue since we found out that our beta channel will be at https://repo.anaconda.cloud/conda-pypi in order to make use of repocore functionality around mirroring. I think we need to figure out an alias solution since it would have been automatic at anaconda.org.

conda_pypi/plugin.py
```diff
@@ -21,9 +21,9 @@ def conda_subcommands():
 @hookimpl
 def conda_post_commands():
     yield CondaPostCommand(
-        name="conda-pypi-ensure-target-env-has-externally-managed",
-        action=ensure_target_env_has_externally_managed,
-        run_for={"install", "create", "update", "remove"},
+        name="conda-pypi-notify-externally-managed-future",
+        action=notify_externally_managed_future,
+        run_for={"install", "create"},
```
jezdez: Should this also warn in case of `conda env create`?

danyeaw: Hi @jezdez, yes, I think that is a good idea!

danyeaw: Hi @jezdez, I updated the PR with this change.

danyeaw:
Hey @jezdez, I pushed some updates:
1. Made some initial docs updates to make it more clear that the feature is the primary one and moving in to beta
2. Added `env_create` to the hook
3. Added some unit tests

I'll approve from my end, but feel free to change or drop any of the changes I made.

danyeaw (approved):

jezdez (commented):
A few notes on the docs changes from the latest commits:

**Channel short name and #16047 dependency**

The quickstart and README now tell users to run `conda config --append channels conda-pypi`. That short name doesn't resolve correctly today because the channel is on `repo.anaconda.cloud`, not `conda.anaconda.org`. This is the exact problem described in conda/conda#16047.

I have draft PRs up for both feedstocks that ship a `condarc.d/conda-pypi.yml` config fragment to fix this:
- AnacondaRecipes: https://github.com/AnacondaRecipes/conda-pypi-feedstock/pull/12
- conda-forge: https://github.com/conda-forge/conda-pypi-feedstock/pull/47

If those land, the short name works automatically when the package is installed. If they don't, the docs here will send users into a broken setup. These two PRs are coupled.

**Uninstallation section**

The new "Uninstallation" section in `quickstart.md` tells users to manually run `conda config --remove channels conda-pypi`. If the `condarc.d` feedstock fix goes through, the channel mapping is removed automatically when the package is uninstalled, so that step becomes unnecessary. The solver/channel_priority cleanup steps would still apply.

**`channel_priority flexible` as a global recommendation**

The README and quickstart both recommend `conda config --set channel_priority flexible` as part of setup. That changes channel priority globally for all operations, not just conda-pypi. This could surprise users who rely on strict priority for their other channels. Worth considering whether this should be documented as a more targeted setting or at least flagged as a tradeoff.

**`env_create` addition**

Adding `env_create` to the `run_for` set in the post-command hook looks correct.

jezdez:
Opened #357 as a draft on top of this branch showing what simplifies if the feedstock `condarc.d` approach lands (conda/conda-pypi#358). Removes the manual channel config steps from docs and the warning message since the channel mapping would ship with the package.

jezdez:
One more thought on the solver and channel_priority settings: while the `condarc.d` approach could technically ship those too (`solver: rattler`, `channel_priority: flexible`), that crosses a line. The channel mapping is passive - it registers a short name that does nothing until someone explicitly uses it. Changing the solver and channel priority would alter default behavior for every conda operation, for every user, just by upgrading conda. That's not a packaging concern anymore, it's a policy change that should go through the normal config/migration path.

danyeaw:
Hi @jezdez, I agree. I think shipping the channel alias is great. I don't think we can change the solver and channel priority on users though.

danyeaw:
I removed most of the docs updates since they are being covered in conda/conda#356.

danyeaw (approved):

PR #359, opened 2026-05-09
Drop conda-rattler-solver dependency

### LINKED ISSUE #350: Drop the conda-rattler-solver dependency
### Checklist

### What is the idea?

Drop the conda-rattler-solver dependency.

### Why is this needed?

#302 fixes support in MatchSpec for `when=` and extra syntax. However, instead updating to the latest conda-rattler-solver release (0.1.0), we should move this dependency to conda so that plugins aren't depending on other plugins.

### What should happen?

Remove the dependency from the recipe.yml and pyproject.toml (commented out dependency). 

### Additional Context

_No response_

jezdez:
## Summary

- Replace the direct `conda-rattler-solver` import (`RattlerSolver`) with `context.plugin_manager.get_cached_solver_backend()`, making conda-pypi solver-agnostic and compatible with whatever solver backend the user has configured (rattler, libmamba, classic).
- Wire both `parse_libmamba_solver_error` and `parse_rattler_solver_error` into `_convert_loop` so wheel discovery handles solver error messages from any backend.
- Remove the `with_rattler_solver` test fixture and let tests run under the default solver backend.
- Remove `conda-rattler-solver` from `pyproject.toml`, `recipe/meta.yaml`, and pixi dependencies.

Closes #350.

## Test plan

- [x] All 139 tests pass locally under the default libmamba solver (2 tests error due to a pre-existing `DualStackServer` socket binding issue in conda's `http_test_server`, unrelated to this change)
- [x] CI passes with the default solver backend
- [x] Verify `test_create_env_from_wheel_channel` and `test_install_from_whl_augmented_repodata` pass in CI (they use the local channel HTTP server which works in CI but not on macOS locally)


danyeaw (approved):
Thanks @jezdez, this makes a nice architectural cleanup.

danyeaw:
@codspeedbot explain why this is faster

PR #368, opened 2026-05-18
Add docstrings to Pixi tasks

pya:
### Description

All tasks in pixi now have docstrings, which will be displayed when running `pixi task list`. This should make it easier for users to understand what each task does and how to use it. All task TOML tables use the long form instead of inline tables, i.e. they are written as:
```toml
[tool.pixi.tasks.task_name]
cmd = ...
```
instead of:

```toml
[tool.pixi.tasks]
task_name = { cmd = "..." }
````

Some tasks where renamed. For example `serve` was renamed to `serve-docs` and `docs` was renamed to `build-docs`. In addition, duplicated commands where replaced with `depends-on`, for example:

```toml
[tool.pixi.feature.test.tasks.test]
...
depends-on = ["dev"]
```

instead of repeating the command from `[tool.pixi.tasks.dev]`.

No functionality was added, removed, or changed. This is purely a documentation update. One exception is:

```toml
[tool.pixi.feature.docs.tasks.make-docs]
depends-on = ["build-docs", "serve-docs"]
```

that adds a new compound command that calls to existing tasks. This is not strictly necessary, but it provides a convenient way to build and serve the docs with a single command.

There no tests for Pixi tasks. So, no new test were added. 

### Checklist

- [- ] Add / update necessary tests? --> no tests for tasks
- [x ] Add / update outdated documentation? --> for tasks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,7 +120,7 @@ sphinx-reredirects = "*"
 sphinx-sitemap = "*"
 
 [tool.pixi.feature.test.tasks.test]
- cmd = 'python -m pytest -m "not benchmark"'
+cmd = 'python -m pytest -m "not benchmark"'
 description = "Run the test suite using pytest, excluding benchmarks. This calls task `dev` first to ensure the environment is set up correctly for testing."
 depends-on = ["dev"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,25 +61,51 @@ unearth = "*"
 [tool.pixi.pypi-dependencies]
 "conda-pypi" = { path  = ".", editable = true }
 
-[tool.pixi.tasks]
-dev = 'python -c "from conda_pypi.python_paths import ensure_externally_managed as e; e()"'
+[tool.pixi.tasks.dev]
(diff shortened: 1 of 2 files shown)


kenodegard:
@conda-bot check

pyproject.toml
```diff
@@ -93,9 +119,16 @@ sphinx-design = "*"
 sphinx-reredirects = "*"
 sphinx-sitemap = "*"
 
-[tool.pixi.feature.test.tasks]
-test = 'python -c "from conda_pypi.python_paths import ensure_externally_managed as e; e()" && python -mpytest -m "not benchmark"'
-benchmark= 'python -c "from conda_pypi.python_paths import ensure_externally_managed as e; e()" && python -mpytest --codspeed'
+[tool.pixi.feature.test.tasks.test]
+ cmd = 'python -m pytest -m "not benchmark"'
```
danyeaw:
```suggestion
cmd = 'python -m pytest -m "not benchmark"'
```

danyeaw (approved):
Thanks @pya, this is a nice improvement!

PR #376, opened 2026-05-20
Add an on/off switch for the pip warnings

### LINKED ISSUE #375: Undesirable warning appearing in 26.5.0
### Checklist

### What happened?

Hello team! I was doing our normal maintenance refresh of DSW and the user-facing image picked up 26.5.0, and started producing this warning:
```
    conda install <package>
```
This is not a warning our users should be seeing, in my opinion, for several reasons:
- The mere existence of pip does not imply someone is intending to pip install a package... unfortunately there are lots of reasons pip gets installed without users' explicit assent
- The apparent mechanism for suppressing the warning seems to be to 1) activate a beta feature and 2) add a channel that the user may or may not have permission to use
- If a user cannot, or will not, be able to add `conda-pypi` to their channels (e.g., firewalled?), they can't follow these instructions
- Many will be surprised and concerned and it is likely to generate support traffic.

Furthermore, I have found that we are now including conda-pypi as a strong dependency of conda 26.5.0, which means that I cannot remove conda-pypi from my user-facing environments without downgrading back to 26.3.2. That's the avenue I am going to to take now, but in my view, I should not have to!

As discussed in Slack, a configuration setting that I can add to a user's conda config file to suppress this warning would work very will for this scenario. I'm hoping that if this warning is not removed, a mechanism for suppressing this warning can be provided.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

travishathaway:
### Description

fixes: #375 

### Checklist - did you ...


travishathaway:
Still need to add documentation and a news file.

conda_pypi/plugin.py
```diff
@@ -39,3 +45,12 @@ def conda_package_extractors():
         extensions=[".whl"],
         extract=extract_whl_as_conda_pkg,
     )
+
+
+@hookimpl
+def conda_settings():
+    yield CondaSetting(
+        name="conda_pypi_pip_warning",
+        description="Enable or disable warning about pip appearing in conda environents",
```
jezdez:
```suggestion
        description="Enable or disable the warning about using pip in conda environents",
```

.github/workflows/docs.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           pixi reinstall --environment docs conda-pypi
           pixi run --environment docs dev
       - name: Build docs
-        run: pixi run docs
+        run: pixi run build-docs
```
jezdez, line 38: Huh, odd!

jezdez (approved):
I think it would be good to mention PEP 668 here as well, LGTM otherwise. Thank you @travishathaway!

.github/workflows/docs.yml
travishathaway: 🤷

conda_pypi/main.py
```diff
@@ -206,16 +206,18 @@ def notify_externally_managed_future(command: str):
...
+    if context.plugins.conda_pypi_pip_warning:
+        logger.warning(
+            "\n"
+            "  This environment has pip installed. A future conda release will\n"
+            "  protect conda environments from accidental 'pip install' usage.\n"
+            "  Try the beta to install PyPI packages natively with conda:\n"
+            "    conda config --set solver rattler\n"
+            "    conda config --append channels conda-pypi\n"
+            "    conda install <package>\n"
+            "  More info: https://docs.conda.io/projects/conda/en/stable/new-features.html"
```
kenodegard, line 219: Include config instructions to disable warning

travishathaway: Good call! Will add 🫡

travishathaway:
@kenodegard,

Mind checking out the updated warning message and leaving an approval if it's okay?

mcg1969:
Thank you _very_ much for doing this!

PR #387, opened 2026-05-28
Refocus pip warning as conda-pypi beta tip

### LINKED ISSUE #385: clarify future warning around pip integrations
I  am getting a lot of questions from my colleagues about this warning messages:

```
    conda install <package>
```

I have a few comments:

1. The warning messages needs to say what would happen if folks did run pip install on a protected env (e.g., add a statement like "Running pip install will result in an error.").
2. A lot of folks use dev installs pip install -e . and local installs pip install .. While it may be obvious to you all, it is not at all obvious to me what will happen with those.
3. The text pointing people to the conda-pypi channel dilutes the important message being displayed. I think it should be removed.

Here are other questions I've gotten:
1. Do I need to change the `pip install .` commands in my conda-forge recipes?
2. The message seems to conflate `pip` and `pypi`. What is happening there? 
3. Claims that "It wants to protect against `pypi` installs (can be confusing) but not against `pip` installs (which is the only way to build packages!)" which I actually do not know the answer to since the externally managed stuff can let installs into some areas go through IIUIC?

cc @travishathaway @jezdez

jezdez:
### Description

Closes #385.

This implements Dan's suggested reframe from the PR discussion: instead of warning users that future conda releases may take direct `pip install` workflows away, the hook now shows a short conda-pypi beta tip at `INFO` level.

The updated message is:

```text
  Did you know? You can install many PyPI packages with conda
  using the conda-pypi beta. Get started:
    https://docs.conda.io/projects/conda/en/stable/new-features.html
```

This also updates the quickstart so the beta docs no longer say `EXTERNALLY-MANAGED` files are automatically added to conda environments during the beta.

cc @mcg1969 @dashagurova @beckermr

### Checklist - did you ...

### DIFF
--- a/conda_pypi/main.py
+++ b/conda_pypi/main.py
@@ -191,8 +191,7 @@ def ensure_target_env_has_externally_managed(command: str):
 
 def notify_externally_managed_future(command: str):
     """
-    Beta-period post-command hook that logs a warning about upcoming
-    EXTERNALLY-MANAGED enforcement instead of placing the marker file.
+    Beta-period post-command hook that points pip users to the conda-pypi beta.
     """
     # Build environments are ephemeral; never show user-facing notices.
     if os.environ.get("CONDA_BUILD_STATE") == "BUILD":
@@ -202,23 +201,17 @@ def notify_externally_managed_future(command: str):
     target_prefix = Path(context.target_prefix)
     if base_prefix == target_prefix or base_prefix.resolve() == target_prefix.resolve():
         return
-    # No point warning about pip protection if pip isn't installed.
+    # No point showing the beta tip if pip isn't installed.
     prefix_data = PrefixData(target_prefix)
     if not list(prefix_data.query("pip")):
         return
 
     if context.plugins.conda_pypi_pip_warning:
-        logger.warning(
+        logger.info(
             "\n"
-            "  This environment has pip installed. A future conda release will\n"
-            "  protect conda environments from accidental 'pip install' usage.\n"
-            "  Try the beta to install PyPI packages natively with conda:\n"
-            "    conda config --set solver rattler\n"
-            "    conda config --append channels conda-pypi\n"
-            "    conda install <package>\n"
-            "  More info: https://docs.conda.io/projects/conda/en/stable/new-features.html"
-            "  To disable this warning, run:"
-            "    conda config --set plugins.conda_pypi_pip_warning false\n"
+            "  Did you know? You can install many PyPI packages with conda\n"
+            "  using the conda-pypi beta. Get started:\n"
+            "    https://docs.conda.io/projects/conda/en/stable/new-features.html\n"
         )
 
 
--- a/conda_pypi/plugin.py
+++ b/conda_pypi/plugin.py
@@ -51,6 +51,6 @@ def conda_package_extractors():
 def conda_settings():
     yield CondaSetting(
         name="conda_pypi_pip_warning",
-        description="Enable or disable the warning about using pip in conda environents",
+        description="Enable or disable the conda-pypi beta tip shown when pip is present",
         parameter=PrimitiveParameter(True),
     )
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -180,11 +180,10 @@ conda pypi install --dry-run niquests pandas
 
 ### Environment protection
 
-`conda-pypi` ships a special file called `EXTERNALLY-MANAGED` that helps
-protect your conda environments from accidental pip usage that could break
-their integrity. This file is automatically installed in the `base`
-environment, all new environments, and existing environments that after running
-a `conda pypi` command on them.
+`conda-pypi` includes support for a special file called `EXTERNALLY-MANAGED`
+that can help protect conda environments from accidental pip usage that could
+break their integrity. During the beta, `conda-pypi` does not automatically add
+this file to conda environments.
 
 More details about this protection mechanism can be found at
 {ref}`externally-managed`.
@@ -196,12 +195,11 @@ More details about this protection mechanism can be found at
 
 #### `conda_pypi_pip_warning`
 
-By default, `conda-pypi` displays a warning when `pip` is detected in a conda
-environment. This warning reminds you that future updates will protect environments
-from using `pip` via the `EXTERNALLY-MANAGED` file defined by
-[PEP 668](https://peps.python.org/pep-0668/).
+By default, `conda-pypi` displays a short beta tip when `pip` is detected in a
+conda environment. The tip points to the conda-pypi beta docs for installing
+PyPI packages with conda.
 
-To disable this warning:
+To disable this tip:
 
 ```bash
 conda config --set plugins.conda_pypi_pip_warning false
--- a/news/385-conda-pypi-beta-tip
+++ b/news/385-conda-pypi-beta-tip
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Refocus the pip warning into a short conda-pypi beta tip. (#385)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -65,7 +65,7 @@ def test_extract_whl_as_conda_pkg(
 
 
 @pytest.mark.parametrize("command", ["install", "create", "env_create"])
-def test_notify_warns_when_pip_installed(
+def test_notify_logs_tip_when_pip_installed(
     mocker: MockerFixture,
     tmp_path: Path,
     command: str,
@@ -80,7 +80,8 @@ def test_notify_warns_when_pip_installed(
 
     notify_externally_managed_future(command)
 
-    mock_logger.warning.assert_called_once()
+    mock_logger.info.assert_called_once()
+    mock_logger.warning.assert_not_called()
 
 
 def test_notify_skips_build_env(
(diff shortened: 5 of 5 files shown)


mcg1969 (commented):
This looks good, thank you! And #376 is also much appreciated

conda_pypi/main.py
```diff
@@ -210,14 +210,17 @@ def notify_externally_managed_future(command: str):
...
-            "    conda install <package>\n"
-            "  More info: https://docs.conda.io/projects/conda/en/stable/new-features.html"
-            "  To disable this warning, run:"
+            "  This environment has pip installed.\n"
+            "  Pip still works in this environment today.\n"
+            "\n"
+            "  In a future conda release, direct pip install commands may fail in\n"
+            "  protected conda environments unless explicitly overridden. This includes\n"
+            "  local installs such as 'pip install .' and editable installs such as\n"
+            "  'pip install -e .'.\n"
```
beckermr: Thanks for clarifying this. 

A few things:

We are telling folks that something they do every day to develop Python packages locally will no longer work. If we actually intend to break that, then we need to give them instructions on how to work around this right here in the warning.

**As far as I am concerned, this is a blocker for ever turning on the externally managed stuff.** These are standard Python dev commands. We want to prevent installs of dev and run dependencies from pypi, not prevent anyone from ever using pip to install a package. Apparently, pip doesn't have a fine-grained enough switch for what we need and so we'll need to work with the Python folks to develop the switch we need.

beckermr (changes requested):

conda_pypi/main.py
dholth: `conda pypi install -e .` exists to install an editable package as a conda package, but it is not fully equivalent to `pip install -e`.

beckermr: Yep, but there is no good reason to break pip completely.

beckermr: What we want, if we want anything at all, is to:

- disallow pip installs from pypi
- turn on the `--no-deps` flag 
- turn on the `-no-build-isolation` flag

Also, how would this change affect users with non-canonical pypi indexes? I am guessing we are breaking things for them too.

beckermr: The docs say 

> With -e / --editable, a local project directory is built into a .conda package and installed.

Assuming I understood the plain english there correctly, this is NOT a replacement for what `-e` does with pip. As I am sure you all know, `-e` with pip ensures that Python imports the code directly from the path it was passed (e.g., `.`). In other words, no copy of the source code is made. (This way one does not have to reinstall the package each time you want to test something with the python interpreter.) 

Did I get this wrong, or did you all really intend to break that?

dholth: It uses https://peps.python.org/pep-0660/ to build an ephemeral "editable wheel" which links to the source checkout; then it converts that to a conda package, installs it, and deletes the ephemeral "editable wheel/conda" package so it can't be accidentally re-used. The build system dependencies come from conda packages, and the equivalent of `--no-build-isolation` and `--no-deps` is used.
We probably want to allow deps in the future, and especially "extras" or named lists of dependencies from Python-land to be installed.
We should improve the short description of what `conda pypi install --editable <path>` does.

beckermr: So what is the difference then? Is it just how it is done on the backend, or something else I am not understanding?

dholth: `conda pypi install -e PROJECT` [uses pypa/build to invoke the project's build system](https://github.com/conda/conda-pypi/blob/main/conda_pypi/build.py#L141) and create a wheel. (Which it converts to a `.conda` to install).
This is better than calling `pip install -e ...` if you want to get all your dependencies from conda repositories, especially because it is able to trap the requirement for `setuptools` or `hatchling` or `flit` etc. and grab those from conda.
The result integrates with the conda environment and conda can upgrade or remove it because it is a conda package.

`conda pypi install <package from pypi>` on the other hand is only able to deal with upstream wheels, and doesn't invoke Python build systems.

beckermr: > This is better than calling pip install -e ... if you want to get all your dependencies from conda repositories, especially because it is able to trap the requirement for setuptools or hatchling or flit etc. and grab those from conda.

I just tested this locally by force uninstall setuptools and trying a dev install. This command appears to make changes to the environment without prompting the user to confirm. That was fine in my case since it simply added a package back. For more complex environments, real damage could be done quickly. It should prompt.

beckermr: xref: https://github.com/conda/conda-pypi/issues/392

beckermr:
```suggestion
            "  In a future conda release, direct pip install commands may fail in\n"
            "  protected conda environments unless explicitly overridden. Failing commands include\n"
            "  local installs such as 'pip install .' and editable installs such as\n"
            "  'pip install -e .'. The commands 'conda pypi install' and \n"
            "  'conda pypi install -e .' may be suitable replacements. Consult\n" 
            "  the 'conda-pypi' documentation for more details.\n"
```

danyeaw:
I think we should refocus the warning from telling users that we will take something away from them (usage of pip) to being a friendly tip (possibly at an INFO level), something like:

```
  Did you know? You can install many PyPI packages with conda
  using the conda-pypi beta. Get started:
    https://docs.conda.io/projects/conda/en/stable/new-features.html
```

Once we have all the pieces in place and both installing from conda-pypi and our editable install commands are at GA, then we can decide if we want to move things to a warning.

beckermr:
The discussion on zulip ended with the conclusion that we'd never enable the externally managed marker in conda envs because it is too breaking.

jezdez:
> The discussion on zulip ended with the conclusion that we'd never enable the externally managed marker in conda envs because it is too breaking.

Can you please copy the thread link about this, I can't find it in the conda Zulip?

Did you also talk about base envs vs normal envs?

beckermr:
I agree `base` envs are a special case. Here is the link: [#general > noisy conda pip notification](https://conda.zulipchat.com/#narrow/channel/457607-general/topic/noisy.20conda.20pip.20notification/with/598411721)

The last message says

> I agree that we have to continue to enable conda users do what they are accustomed to even if "better" workflows emerge.

I interpreted this as meaning that breaking `pip` commands in standard conda envs is not something we want to do.

beckermr (approved):

jezdez:
I've updated the log message following what was discussed here, I need sign off from: @danyeaw @mcg1969 @beckermr @dholth.

As the person who has participated in PEP668 in 2021, I'm very bullish to ship the ability to protect conda users from pip messing things up. It seems we haven't found community acceptance, though, given the limited data presented here, which is important to not ignore. I absolutely think that "hobbling users" is the other side of the story of "providing footguns". Using pip in conda environments is a very common footgun.

mcg1969:
I appreciate the opportunity to weigh in — and at the same time acknowledge my voice is but one of many.

I totally agree it's a common footgun. Our support team has a significant amount of experience dealing with people with big ol' holes in their feet. And yet... in many cases, it's a risk they have no choice but to take.

The single most important way to protect conda users from pip messing things up is to provide as many conda-installable packages as possible. I chose "conda-installable" deliberately, because conda-pypi is a great solution to better manage the use of wheels inside conda envionments. Its heart is in the right place. My prior objections to the messaging have been _largely_ addressed and this PR is a further step in the right direction.

I do still question whether or not we should be offering this messaging when `conda-pypi` is still in beta; I would have preferred we wait until we could declare it to be GA. And I also wonder if using the already-existing conda-notices function would have been a better way to promote it, for a time.

So the tl;dr: I definitely prefer this messaging over what is currently baked in. I remain skeptical that there should be any such warning _at this stage_. But with the configuration change making it possible to disable it, I don't need to assert my position too strongly; my customers can work around it.

danyeaw (approved):
Thanks @jezdez! I think this is a big improvement.

mcg1969 (approved):

beckermr:
Thanks @jezdez!

> As the person who has participated in PEP668 in 2021, I'm very bullish to ship the ability protect conda users from pip messing things up. It seems we haven't found community acceptance though given the limited data presented here, which is important to not ignore. I absolutely think that "hobbling users" is the other side of the story of "providing footguns". Using pip in conda environments is a very common footgun.

I understand your point here and appreciate it. I also appreciate your work on the PEP as well. I know my comments on this issue were strong, and my apologies again for that. I did not properly recognize all the hard work you all do. Thank you!

My intuition here for not breaking existing workflows has been motivated and sharpened by the ongoing turnover in Python packaging tooling (which has settled somewhat, to be fair). Some stuff there is still broken and/or not working in the fallout, years later. A good example of this is the ongoing `--config-settings` saga with setuptools (see https://github.com/pypa/setuptools/issues/2491, https://github.com/pypa/setuptools/discussions/4083, https://github.com/pypa/setuptools/issues/3896#issuecomment-1656714771). Things are still not quite working, and the tools deprecated things too quickly, IMHO, without clear, adequate replacements. 

An example of where `conda-pypi` cannot replace generic `pip` usage is in installation instructions for some HPC environments. These sometimes have folks install non-src or non-pure-python packages because the versions on conda/conda-forge do not "speak" properly to the underlying system libraries. That is ultimately a conda/conda-forge problem, but breaking the escape hatch for the workaround is destructive until we can provide better replacements and workflows. A possible approach here is repackaging the binary wheel with a full conda-build using special flag, if only to make it easier to remove it later.

I 100% agree `base` environments need to be protected, and I think that should be forcibly rolled out as soon as possible. I am pretty sure either conda-pypi or conda-self just did this in the latest release of conda. I had some `pypi` craziness in my own `base` env, and the latest conda versions helped me clean it up. That was hugely helpful.

jezdez:
Thanks all for your feedback and willingness to work on improving things for users, this means a lot even if it takes effort to get there sometimes!

PR #400, opened 2026-06-03
Split up atomic `find_and_fetch` method into two parts via PEP 658/714

agriyakhetarpal:
### Description

This is a small refactor of the `find_and_fetch` method we have in `conda_pypi/downloader.py`. My rationale is that we can use the package repository metadata layout as defined in [PEP 658](https://peps.python.org/pep-0658/) and implemented in PyPI to split this function into two phases: one that resolves the link to the wheel on PyPI (`_find_wheel_link`), and a second phase that downloads the wheel for further processing and all other uses (`_download_wheel`).

In this case, we can now use the split to use the PEP 658 primitives I have added here. This PR does not implement such usage; that is something I have done in #312 for implementing support for PEP 794. The idea there is that we then do: (i) find the wheel's link, (ii) download the wheel's METADATA and check it for conflicts between `Import-Name` and `Import-Namespace` fields, and (iii) download the wheels. This is much more performant and saves bandwidth than proceeding to download entire wheels and detecting PEP 794 metadata conflicts _after_ doing so.

I have defined a `fetch_pep658_wheel_metadata` function here but have not **intentionally** used it yet, as it doesn't make sense to do so just now. I have added a few tests for it, though – which are just guards and basic functionality checks.

I can make `_find_wheel_link` and `_download_wheel` public if it is suggested for me to do so. `find_and_fetch` is no longer used but for two pieces of test code, and I'd suggest that we put it up for deprecation in a follow-up PR after this (or I can do so now, if suggested).

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-pypi/blob/main/news/TEMPLATE)) for the next release's release notes?~ This is just an internal refactor and has no public API changes.
     

agriyakhetarpal:
What this PR enables can be better seen in this commit over at #312: [`6cf3a9a`](https://github.com/conda/conda-pypi/pull/312/changes/6cf3a9aa74cf0882699f541758cd4899bb38136c). `ConvertTree._convert_loop` has three phases, and the second phase is where we call `_check_pep794_conflicts_pre_download` (which is possible due to the PEP 658 metadata fetching enabled here).

I have not removed `find_and_fetch` here; instead, I kept it as a convenience wrapper, since it is a public function and we'd probably need to go through a deprecation period to do so.

conda_pypi/downloader.py
```diff
@@ -51,25 +51,64 @@ def find_package(finder: PackageFinder, package: str):
...
         raise CondaPypiError(
             f"No wheel file available for {package}. "
             f"Only source distributions are available. "
             f"conda-pypi requires wheel files for conversion."
         )
+    return link
 
-    log.info(f"Fetch {package} as {filename}")
+
+def _download_wheel(link, target: Path) -> Path:
```
jezdez, line 99: Please ask the AI to use fewer private functions; there is no need for this

agriyakhetarpal: Hi @jezdez, thanks for reviewing – could you please elaborate? I did not steer AI for drafting this PR, even though I used it for the companion PEP 794 PR. Do you feel it would be better to leave these functions public? I do not wish to unintentionally add any public API pieces unless it's really needed.

jezdez: Thanks for clarifying. My concern is not that these need to become a supported public API, but that if #312 imports them from another module, underscored helper names become misleading. I’d prefer either making them regular downloader module helpers (`find_wheel_link` / `download_wheel`, undocumented is fine), or keeping the higher-level resolve/check/download workflow inside `downloader.py`.

Also, for the PEP 658 helper, can we pass the `unearth.Link` and use `link.dist_info_link` rather than accepting a raw URL and appending `.metadata`? Unearth already tracks whether metadata was advertised by the index.

PR #401, opened 2026-06-05
Prompt before editable install changes

### LINKED ISSUE #392: prompt user if env changes are being made
The `conda pypi install -e .` command made changes to my environment without prompting me for a yes or no. This could cause real damage quickly and the command should prompt.

cc @dholth

jezdez:
### Description

Fixes #392.

`conda pypi install -e .` could change the target environment without giving the user the normal yes/no gate. This makes editable installs follow the same prompt controls as the rest of `conda pypi install`.

This PR:

- adds editable `--dry-run` support, including JSON output
- passes `--yes` through build-requirement installs and the final editable package install
- prompts before the ephemeral editable conda package is installed when `--yes` is not set
- supports repeated `-e` / `--editable` for installing multiple local projects in order
- keeps editable project paths and normal package specs separate for now
- updates the editable-install docs and release note

Validation:

- `pixi run --environment test-py313 python -m pytest -q`
- `pixi run --environment lint ruff check .`
- `pixi run --environment lint ruff format --check conda_pypi/cli/install.py conda_pypi/build.py conda_pypi/dependencies/pypi.py conda_pypi/installer.py tests/cli/test_install.py tests/test_editable.py`

### Checklist - did you ...

### DIFF
--- a/conda_pypi/build.py
+++ b/conda_pypi/build.py
@@ -107,6 +107,7 @@ def build_pypa(
     output_path,
     prefix: Path,
     distribution="editable",
+    yes: bool = True,
 ):
     """
     Args:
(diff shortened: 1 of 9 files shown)


dholth (approved):
YES

beckermr (approved):
I am a n00b on this level of dev on the CLI. From what I can see, LGTM, but take that with a grain of salt.

Thank you!

PR #408, opened 2026-06-09
Remove post install plugin that uses pip to install packages

soapy1:
### Description

Instead users should use the `conda-pypi` channel with the conda-pypi plugin in order to install packages. ref: https://conda.github.io/conda-pypi/quickstart/

### Checklist - did you ...

- ~[ ] Add / update necessary tests?~
- ~[ ] Add / update outdated documentation?~


conda_pypi/post_command/install.py
```diff
@@ -1,124 +0,0 @@
...
-                    elif on_error == "error":
-                        errors.append(msg[0] % msg[1:])
-                    else:
-                        log.debug(*msg)
-    if errors:
-        errors = "\n- ".join(errors)
-        raise CondaVerificationError(f"PyPI packages checksum verification failed:\n- {errors}")
-
-
-def post_command(command: str):
```
soapy1, line 78: Does this need to go thru a formal deprecation process?

ryanskeith: My vote is no we do not as the code still is in beta according to its version 0.9.0.

ryanskeith: Ok, this is a breaking change after looking at it closely.  What we lose is lockfile support for pypi.  Any lockfiles out there are now orphaned.  I am not sure if this will break things outright or just skipped.   We should probably ask a few more people.  This makes me less confident about my previous thought.

conda_pypi/post_command/install.py
```diff
@@ -1,124 +0,0 @@
...
-
-    pypi_lines = _pypi_lines_from_paths()
-    if not pypi_lines:
-        return 0
-
-    with get_spinner("\nPreparing PyPI transaction"):
-        pkgs = _prepare_pypi_transaction(pypi_lines)
-
-    with get_spinner("Executing PyPI transaction"):
-        run_pip_install(
```
ryanskeith, line 90: I think this may be the only place `run_pip_install` is used.  If we do remove it, we should probably remove it with all other dependent functions for a clean code removal.  I think there are a few more that fall in this bucket.

ryanskeith (commented):
This removal needs a few more eyes as this is a breaking change (albeit maybe a good breaking change) for conda-pypi users.

Also, if the hook is removed, we should also remove any orphaned code.

conda_pypi/post_command/install.py
soapy1: hmmm, which lockfile format uses this?

This post command action only applies to lines that are prefixed with `# pypi: `. conda-lock-v1 format [does not include pip packages in it's lockfile output ](https://github.com/conda/conda-lockfiles/blob/main/conda_lockfiles/conda_lock/v1.py#L198). It also looks like this is not included in the rattler type lockfile.

I think conda-pypi should probably not be sharing this type of logic/have these types of file reading responsibilities. It's probably good for conda pypi to be decoupled from needing to understand any environment file formats.

conda_pypi/post_command/install.py
soapy1: Good idea! Are there any other things that depend on this plugin that are using this? I took a look and couldn't find anything.

jezdez (approved):

conda_pypi/post_command/install.py
soapy1: after discussing this offline, it looks like folks are mostly happy with just removing this piece of code. Nothing should be relying on this `# pypi: ` marker.

@ryanskeith does that sound like an ok resolution to you?

ryanskeith: If people are happy I am fine.  I may have mispoke here and meant environment files.

PR #410, opened 2026-06-11
Add `index` subcommand

ForgottenProgramme:
### Description

resoves #278 

Implements the `conda pypi index` subcommand as described in #278.
This subcommand allows users to create a local conda channel from a directory of pure Python wheel files, generating a repodata.json that conda's solver can use to resolve and install wheel-based packages.

**Changes**
- Added conda_pypi/cli/index.py with the index subcommand implementation
- Added relevant tests

Added conda_pypi/cli/index.py with the index subcommand implementation

### Checklist - did you ...


ForgottenProgramme:
Hey, @danyeaw !
We are expecting the directory structure to be like this:

```
<root>/
  <package_name_1>/
    <package_name_1>-<version>-<tags>.whl

  <package_name_2>/
    <package_name_2>-<version>-<tags>.whl

  <package_name_3>/
    <package_name_3>-<version>-<tags>.whl
 
```
right?

danyeaw:
Hi @ForgottenProgramme, that's correct. Going back to the use case for this feature: someone hosting PyPI simple index or mirror needs a way to generate conda repodata from .whl files on disk. What you described is the [Manual Index](https://packaging.python.org/en/latest/guides/hosting-your-own-index/#manual-repository) directory structure, so that is what we want to support initially for this feature.

ForgottenProgramme:
@danyeaw , if there are invalid entries in the "root" directory (i.e. anything other than subdirectories with wheel package names), should we abort the indexing process or just ignore those entries and carry on?

danyeaw:
Hi @ForgottenProgramme, I think we should just ignore everything that isn't wheel filenames in the directories. It would be common that next to the wheels would be source distributions of packages.

ForgottenProgramme:
@danyeaw, 
I meant at the root level. For example, if there are files other than expected wheel directories. i.e. Anything other than `<package_name_1>/`, `<package_name_2>/` ...  in the `root/`.

danyeaw:
@ForgottenProgramme, I would ignore those as well. We should avoid erroring out unless there is really some type of conflict we can't handle.

ForgottenProgramme:
> @ForgottenProgramme, I would ignore those as well. We should avoid erroring out unless there is really some type of conflict we can't handle.

I agree!

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,145 @@
...
+    index = parser.add_parser(
+        "index",
+        help=summary,
+        description=description,
+        epilog=epilog,
+    )
+    index.add_argument(
+        "directory",
+        metavar="DIRECTORY",
+        type=Path,
```
dholth, line 36: It's possible to pass a callable here.

```
...
type=existing_directory
...

def existing_directory(directory: str):
    path = Path(directory)
    if not path.is_dir():
        raise SomeException()
    return path
```

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def pypi_data_dict(wheel: Path, wheel_metadata: PackageMetadata):
+    """Return expected dict structure of pypi metadata with the relevant fields"""
+
+    # convert to json format using the `json` property of the PackageMetadata class
+    wheel_metadata_json= wheel_metadata.json
+    # generate the expected dict structure of pypi metadata with the relevant fields as needed by the `pypi_to_repodata` function.
+    pypi_data = {
+            "info": {
```
dholth: Please configure your editor to format-on-save; and install pre-commit and run `pre-commit install` so it will automatically lint and format before each commit. pixi tasks are also available to lint and format.

danyeaw (commented):
This is just a preliminary PR review (I didn't look at the code in detail). Things that look amazing:

1. CLI changes to add the subcommand
2. The general end-to-end shape is there
3. `pypi_data_dict()` is the right converter, nice
4. `validate_dir_and_return_whl_files()` is a nice helper

What could still use more work:
1. You modified `update_index()`, but make sure you update the other users of it as well
2. `create_channel_index()` is missing wheel repodata (v3) config (reference `generate_noarch_wheel_repodata.py`
3. The url in `pypi_data_dict()` is pointing to the wheel, but I think it needs the base_url/directory/wheel name

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def configure_parser(parser: _SubParsersAction) -> None:
+    """Configure all subcommand arguments and options via argparse"""
+
+    summary = "Index a directory of `.whl` files to generate repodata.json"
+    description = summary
+    epilog = dals("""
+    Generate a `repodata.json` file from a directory of `.whl` files.
+    This is useful for creating a local conda channel from a collection of wheel files.
```
dholth:
```suggestion
    Creates a local conda channel from a collection of wheel files.
```

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,145 @@
...
+            "name": wheel_metadata_json.get("name"),
+            "version": wheel_metadata_json.get("version"),
+            "requires_dist": wheel_metadata_json.get("requires_dist", []),
+            "requires_python": wheel_metadata_json.get("requires_python"),
+        },
+        "urls": [
+            {
+                "packagetype": "bdist_wheel",
+                "filename": wheel.name,
+                "url": str(wheel),
```
danyeaw: Hey @ForgottenProgramme, we discussed this when we met just now, but I thought I would write it down as well for your reference. I recommend that we add a `--base-url` CLI option so users can pass in a server URL (like https://packages.mahe.org/). The reference for this is CEP-15. If it isn't provided, then the current directory passed in can be set as the URI in the `file://` format. I don't think the CEP specifies if base-url has a trailing slash or not, so we probably need to normalize that.

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,145 @@
...
+                )
+        except Exception as e:
+            print(f"Failed to process {wheel}: {e}")
+            failed_wheels.append(wheel)
+            continue
+
+        pypi_data = pypi_data_dict(wheel, wheel_metadata)
+
+        # store the converted metadata in the conda index cache
+        store_pypi_metadata(cache, pypi_data)
```
danyeaw:
```suggestion
        try:
            with WheelFile.open(wheel) as source:
                wheel_metadata = package_metadata_from_metadata_body(
                    source.read_dist_info("METADATA")
                )
            pypi_data = pypi_data_dict(wheel, wheel_metadata)
            store_pypi_metadata(cache, pypi_data)
        except UnableToConvertToRepodataEntry as e:
            print(f"Skipping {wheel.name}: {e}")
            failed_wheels.append(wheel)
        except InvalidRequirement as e:
            print(f"Skipping {wheel.name}: invalid metadata ({e})")
            failed_wheels.append(wheel)
        except ValueError as e:
            print(f"Skipping {wheel.name}: {e}")
            failed_wheels.append(wheel)
        except (OSError, zipfile.BadZipFile) as e:
            print(f"Failed to read {wheel.name}: {e}")
            failed_wheels.append(wheel)
```

I would recommend moving `pypi_data_dict` and `store_pypi_metadata` inside the try/except and catch the more specific errors that they could raise. Otherwise if you have a bad wheel, a compiled wheel, or bad metadata, the whole indexing will fail. Adding a few tests for these error cases would be good as well.

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,145 @@
...
+    }
+    return pypi_data
+
+
+def execute(args: Namespace) -> int:
+    """Entry point for the `conda pypi index` subcommand"""
+
+    directory = args.directory
+    # ensure directory is provided
+    if not directory:
```
danyeaw: For CLI usage, argparse will always ensure that `args.directory` exists, so I don't think this check or the test of it is needed, because we are testing argparse itself.

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,145 @@
...
+            }
+        ],
+    }
+    return pypi_data
+
+
+def execute(args: Namespace) -> int:
+    """Entry point for the `conda pypi index` subcommand"""
+
+    directory = args.directory
```
danyeaw:
```suggestion
    directory = Path(args.directory).expanduser()
```

This would match what `install.py` is doing for consistency. This allows someone to put in `~/mywheels` and it would expand to the full directory.

danyeaw (commented):
Hey @ForgottenProgramme, this is coming along great - I can really see the nice architecture of the wheels pipeline coming through now. I just had a few comments about allowing for a base_url and further tweaks to the error handling to make the command more robust to errors from individual wheels.

conda_pypi/cli/index.py
ForgottenProgramme: Done! :)

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,158 @@
...
+            {
+                "packagetype": "bdist_wheel",
+                "filename": wheel.name,
+                "url": base_url + wheel.name,
+                "size": wheel.stat().st_size,
+                "digests": {"sha256": sha256_checksum(str(wheel))},
+            }
+        ],
+    }
+    return pypi_data
```
jezdez: The generated URL drops the package subdirectory from the supported `<root>/<package>/<wheel>` layout. Build it from the wheel path relative to the indexed root so the default `file://` URL and any supplied base URL point at the actual wheel.

```suggestion
def pypi_data_dict(
    root: Path, wheel: Path, wheel_metadata: PackageMetadata, base_url: str
):
    """Return expected dict structure of pypi metadata with the relevant fields"""

    wheel_metadata_json = wheel_metadata.json
    pypi_data = {
        "info": {
            "name": wheel_metadata_json.get("name"),
            "version": wheel_metadata_json.get("version"),
            "requires_dist": wheel_metadata_json.get("requires_dist", []),
            "requires_python": wheel_metadata_json.get("requires_python"),
        },
        "urls": [
            {
                "packagetype": "bdist_wheel",
                "filename": wheel.name,
                "url": base_url + wheel.relative_to(root).as_posix(),
                "size": wheel.stat().st_size,
                "digests": {"sha256": sha256_checksum(str(wheel))},
            }
        ],
    }
    return pypi_data
```

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,158 @@
...
+    channel_index = create_channel_index(directory)
+    cache = channel_index.cache_for_subdir("noarch")
+
+    for wheel in all_wheels:
+        try:
+            with WheelFile.open(wheel) as source:
+                wheel_metadata = package_metadata_from_metadata_body(
+                    source.read_dist_info("METADATA")
+                )
+            pypi_data = pypi_data_dict(wheel, wheel_metadata, base_url)
```
jezdez: Pass the indexed root through so `pypi_data_dict()` can keep package subdirectories in generated URLs.

```suggestion
            pypi_data = pypi_data_dict(directory, wheel, wheel_metadata, base_url)
```

conda_pypi/index.py
```diff
@@ -6,23 +6,29 @@
...
-def update_index(path):
+def create_channel_index(path):
     channel_index = ChannelIndex(
         path,
         None,
         threads=1,
-        debug=False,
-        write_bz2=False,
         write_zst=True,
         write_run_exports=True,
```
jezdez: This wheel-only index path does not need run exports. Leaving this on prints one `doesn't look like a conda package` warning per valid wheel during successful indexing.

```suggestion
        write_run_exports=False,
```

conda_pypi/cli/index.py
ForgottenProgramme: hmm. 
that only works correctly when the `--base-url` is passed. Otherwise, `base_url` is already the he `file://` URI of `root `

ForgottenProgramme: I solved this by generating the url before the dict is generated. The url is created different based on whether the `--base-url` flag was passed or not. 

```
if args.base_url:
  url = base_url + wheel.relative_to(directory).as_posix()
else:
  url = wheel.resolve().as_uri()
```

conda_pypi/index.py
ForgottenProgramme: removing it completely, since it defaults to `False` anyways

tests/cli/test_index.py
```diff
@@ -0,0 +1,149 @@
...
+    pkg_dir = tmp_path / "bad-package"
+    pkg_dir.mkdir()
+    make_wheel(pkg_dir, "bad_package", "1.0.0", platform="cp311-win_amd64")
+
+    args = Namespace(directory=tmp_path, base_url=None)
+    result = execute(args)
+
+    assert result == 0
+    captured = capsys.readouterr()
+    assert "not a pure-python whee" in captured.out
```
danyeaw:
```suggestion
    assert "not a pure-python wheel" in captured.out
```

Nitpick for a typo.

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,159 @@
...
+            }
+        ],
+    }
+    return pypi_data
+
+
+def execute(args: Namespace) -> int:
+    """Entry point for the `conda pypi index` subcommand"""
+
+    directory = Path(args.directory).expanduser()
```
danyeaw, line 112: Minor / not-blocking, might be nice to add an expanduser test.

tests/cli/test_index.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    out, err, rc = conda_cli("pypi", "index", "--help", raises=SystemExit)
+    assert rc.value.code == 0
+    assert "DIRECTORY" in out
+
+
+def test_validate_dir_not_a_directory(tmp_path):
+    """Test invalid dir"""
+    not_a_dir = tmp_path / "file.txt"
+    with pytest.raises(Exception):
```
danyeaw:
```suggestion
    with pytest.raises(ArgumentError, match="Not a directory"):
```

It would strengthen this test to catch the ArgumentError now that we have more precise exception handling.

tests/cli/test_index.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+
+import json
+import shutil
+import zipfile
+from argparse import Namespace
+from pathlib import Path
+
+import pytest
+
```
danyeaw, line 13:
```suggestion
import pytest
from conda.exceptions import ArgumentError

```

Needed for the suggested change below.

conda_pypi/cli/index.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+    index.add_argument(
+        "directory",
+        metavar="DIRECTORY",
+        type=Path,
+        help="Directory containing .whl files to index",
+    )
+    index.add_argument(
+        "--base-url",
+        help="Base URL for the channel (e.g. https://packages.example.com/). Defaults to a file:// URI of the given directory.",
```
danyeaw:
```suggestion
        help="Base URL for the channel (e.g. https://packages.example.com/). When omitted, each entry uses a file:// URI for each wheel file.",
```

Since the URI points to an absolute path of each wheel, I think this is more accurate than saying it points to the directory.

danyeaw (commented):
Hey @ForgottenProgramme, this is looking great!

We should probably have some user facing docs (reference commands, features, mention in index, and in the quickstart. Would you prefer to do that in a separate PR?

danyeaw (approved):

PR #420, opened 2026-06-18
Derive wheel build string for index.json from WHEEL metadata

### LINKED ISSUE #417: Wheel extractor should derive build string from wheel tags, not hardcode pypi_0
### What is the idea?
The .whl extractor should write `build=py3_none_any_0` in index.json, derived from the wheel filename tags, rather than hardcoding `pypi_0`.

### Why is this needed?
During a normal solver install, the solver record overrides index.json and the correct metadata is written to conda-meta (e.g. pyqplot-0.7.2-py3_none_any_0.json with `build=py3_none_any_0`, `channel=conda-pypi/noarch`). During lockfile restore, only index.json is available, so the hardcoded pypi_0 ends up in the prefix record. Partially fixes https://github.com/conda/conda/issues/16217.

### What should happen?
The extractor derives the build string from wheel tags so that index.json matches the repodata v3 channel record.

### Additional Context
A companion conda fix (_get_json_fn filename naming) is also needed since the URL basename and build string can still differ even after this fix.

ryanskeith:
### Description

Fixes #417.

- Added parsing code to look at WHEEL meta data for wheel packages.
- Opinionated choices when more than one tag is present.
- Leverage build_number if provided and conda can use it (aka pure numeric).
- Updated tests.

### Checklist - did you ...

### DIFF
--- a/conda_pypi/package_extractors/whl.py
+++ b/conda_pypi/package_extractors/whl.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 from email.parser import HeaderParser
 from logging import getLogger
--- a/conda_pypi/package_extractors/whl.py
+++ b/conda_pypi/package_extractors/whl.py
@@ -5,7 +5,7 @@ from email.parser import HeaderParser
 from logging import getLogger
 from os import PathLike
 from pathlib import Path
-from typing import BinaryIO, Iterable, Literal, Tuple, TYPE_CHECKING
+from typing import TYPE_CHECKING, BinaryIO, Iterable, Literal, Tuple
 
 import installer.utils
 from installer.destinations import WheelDestination
@@ -31,6 +31,7 @@ SCHEME_TO_CONDA_PREFIX: dict[Scheme, str] = {
 
 if TYPE_CHECKING:
     from email.message import Message
+
 
 def _create_build_string_from_wheel_meta_and_filename(
     wheel_meta: Message,
@@ -66,6 +67,7 @@ def _create_build_string_from_wheel_meta_and_filename(
 
     build_string = f"{wheel_tag.replace('-', '_')}_{build_number}"
     return build_string, build_number
+
 
 # inline version of
 # from conda.gateways.disk.create import write_as_json_to_file
@@ -174,7 +176,9 @@ class MyWheelDestination(WheelDestination):
 
         wheel_meta = HeaderParser().parsestr(source.read_dist_info("WHEEL"))
 
-        build_string, build_number = _create_build_string_from_wheel_meta_and_filename(wheel_meta, wheel_filename)
+        build_string, build_number = _create_build_string_from_wheel_meta_and_filename(
+            wheel_meta, wheel_filename
+        )
 
         index_json_data = {
             "name": package_name,
--- a/conda_pypi/package_extractors/whl.py
+++ b/conda_pypi/package_extractors/whl.py
@@ -36,8 +36,14 @@ def write_as_json_to_file(file_path, obj):
 
 
 class MyWheelDestination(WheelDestination):
-    def __init__(self, target_full_path: str | Path, source: WheelFile):
+    def __init__(
+        self,
+        target_full_path: str | Path,
+        source: WheelFile,
+        whl_full_path: str | Path,
+    ):
         self.target_full_path = Path(target_full_path)
+        self.whl_full_path = Path(whl_full_path)
         self.sp_dir = self.target_full_path / "site-packages"
         self.entry_points = []
         self.source = source
@@ -122,10 +128,10 @@ class MyWheelDestination(WheelDestination):
         }
         write_as_json_to_file(info_dir / "paths.json", paths_json_data)
 
-        # index.json — fn from zip path; Tag and Build from WHEEL dist-info.
+        # index.json — fn from wheel path; Tag and Build from WHEEL dist-info.
         package_name = str(source.distribution)
         package_version = str(source.version)
-        wheel_filename = Path(source._zipfile.filename).name
+        wheel_filename = self.whl_full_path.name
 
         wheel_meta = HeaderParser().parsestr(source.read_dist_info("WHEEL"))
         tag_lines = wheel_meta.get_all("Tag") or []
@@ -191,9 +197,10 @@ class MyWheelDestination(WheelDestination):
 
 
 def extract_whl_as_conda_pkg(whl_full_path: str | Path, target_full_path: str | Path):
+    whl_full_path = Path(whl_full_path)
     with WheelFile.open(whl_full_path) as source:
         installer.install(
             source=source,
-            destination=MyWheelDestination(target_full_path, source),
+            destination=MyWheelDestination(target_full_path, source, whl_full_path),
             additional_metadata={"INSTALLER": b"conda-via-whl"},
         )
--- a/conda_pypi/package_extractors/whl.py
+++ b/conda_pypi/package_extractors/whl.py
@@ -1,5 +1,6 @@
 import json
 from email.parser import HeaderParser
+from logging import getLogger
 from os import PathLike
 from pathlib import Path
 from typing import BinaryIO, Iterable, Literal, Tuple
@@ -13,6 +14,8 @@ from packaging.tags import parse_tag
 
 from conda_pypi.license_files import copy_into_info_licenses, package_metadata_from_metadata_body
 from conda_pypi.utils import sha256_base64url_to_hex
+
+logger = getLogger(__name__)
 
 # Maps wheel scheme names to their conda package directory prefix.
 # An empty string means files go directly under the package root (env prefix).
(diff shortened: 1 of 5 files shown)


conda_pypi/package_extractors/whl.py
```diff
@@ -117,21 +119,36 @@ def _create_conda_metadata(
...
+        tag_lines = wheel_meta.get_all("Tag") or []
+        # WHEEL lists expanded tags; filenames may compress (py2.py3-none-any). Expand via
+        # parse_tag when needed and prefer py3-none-any to match repodata v3.
+        if tag_lines:
+            tag_candidates = tag_lines
+        else:
+            # parse_tag will expand the compressed tag format (py2.py3-none-any) to a list of tags.
+            tag_candidates = [str(t) for t in parse_tag(parse_wheel_filename(wheel_filename).tag)]
+        # Prefer py3-none-any to match repodata v3.
+        wheel_tag = next((t for t in tag_candidates if t == "py3-none-any"), tag_candidates[0])
```
ryanskeith: Assuming here that the 1st one the best.  Maybe we could improve the logic?

conda_pypi/package_extractors/whl.py
```diff
@@ -117,21 +119,36 @@ def _create_conda_metadata(
...
+
+        build_number = 0
+        wheel_build = wheel_meta.get("Build")
+        if wheel_build is not None and wheel_build.isdigit():
+            build_number = int(wheel_build)
```
ryanskeith: I am not sure if this is worth the time to get the wheel_build number.  Seems like low hanging fruit though.

conda_pypi/package_extractors/whl.py
```diff
@@ -117,21 +119,36 @@ def _create_conda_metadata(
...
+
+        wheel_meta = HeaderParser().parsestr(source.read_dist_info("WHEEL"))
+        tag_lines = wheel_meta.get_all("Tag") or []
+        # WHEEL lists expanded tags; filenames may compress (py2.py3-none-any). Expand via
+        # parse_tag when needed and prefer py3-none-any to match repodata v3.
+        if tag_lines:
+            tag_candidates = tag_lines
+        else:
+            # parse_tag will expand the compressed tag format (py2.py3-none-any) to a list of tags.
+            tag_candidates = [str(t) for t in parse_tag(parse_wheel_filename(wheel_filename).tag)]
```
ryanskeith: The code is falling back to wheel file name which should be accurate in the highest majority of cases.  I do not like that the code needs to use `source._zipfile.filename` though.  An alternative could be to pass the filename to the `_create_conda_metadata` function call...  This probably would mean the class API would need to change thus this this route was chosen.

conda_pypi/package_extractors/whl.py
ForgottenProgramme: Maybe we can still choose the highest Python version instead of just choosing the first tag?

conda_pypi/package_extractors/whl.py
```diff
@@ -117,21 +119,36 @@ def _create_conda_metadata(
...
-        # The extension is required because _get_json_fn() uses endswith() to detect package type
+        # index.json — fn from zip path; Tag and Build from WHEEL dist-info.
         package_name = str(source.distribution)
         package_version = str(source.version)
-        build_string = "pypi_0"
-        fn = f"{package_name}-{package_version}-{build_string}.whl"
+        wheel_filename = Path(source._zipfile.filename).name
+
+        wheel_meta = HeaderParser().parsestr(source.read_dist_info("WHEEL"))
+        tag_lines = wheel_meta.get_all("Tag") or []
```
ForgottenProgramme: We already have `_parse_wheel_file` and `_tags_from_wheel` static methods defined in `PyPIDistribution` class. Since these are private to that class, calling them from here would create unwanted coupling. Maybe we should move them into `conda_pypi/utils.py` and use them from both places?

ryanskeith: Thanks for finding this commonality.  As I looked into it I feel the WHEEL parsing overlap is superficial. `_tags_from_wheel` decomposes tags into abi/platform for lockfile reconstruction, while the implemented code picks a single tag. `_parse_wheel_file` is also path-based, whereas the extractor reads WHEEL from an open WheelFile.   I feel for this PR this is the right fit.  If another use case pops up, I think that would be the time to look at common functionality.

tests/test_conda_meta_json_filename.py
```diff
@@ -1,21 +1,46 @@
...
+    [
+        ("py3-none-any", "py3-none-any"),
+        ("py2-none-any", "py3-none-any"),
+        ("py38-none-any", "py3-none-any"),
+        ("cp312-cp312-win_amd64", "cp312-cp312-win_amd64"),
+    ],
+)
+def test_noarch_wheel_tag_normalization_for_index_build(wheel_tag, expected):
+    normalized = wheel_tag
+    if normalized != "py3-none-any" and normalized.endswith("-none-any"):
```
danyeaw: I don't think this test is exercising `extract_whl_as_conda_pkg`. It would also be nice to add changes to the build number in to the parameterized test as well.

conda_pypi/package_extractors/whl.py
danyeaw: Could we add a `whl_full_path` property to the `MyWheelDestination` class instead of using the private method?

danyeaw (commented):
Hey @ryanskeith, thanks for working on this, this will be a huge way towards getting lockfiles fully functional with conda-pypi. Overall this looks like a great implementation 👍

conda_pypi/package_extractors/whl.py
ryanskeith: Sounds good.

tests/test_conda_meta_json_filename.py
ryanskeith: Ugh, it seems I didn't actually hook that test up properly.

tests/test_conda_meta_json_filename.py
```diff
@@ -9,26 +9,80 @@
...
 
 import json
+import zipfile
 from pathlib import Path
 
 import pytest
-from installer.sources import WheelFile
+
+
+def _make_wheel(
```
danyeaw: Hey @ryanskeith, for performance instead of creating these new wheels every time the test is run, why don't we save the test wheels to the repo?

conda_pypi/package_extractors/whl.py
```diff
@@ -36,8 +36,14 @@ def write_as_json_to_file(file_path, obj):
 
 
 class MyWheelDestination(WheelDestination):
-    def __init__(self, target_full_path: str | Path, source: WheelFile):
+    def __init__(
+        self,
+        target_full_path: str | Path,
+        source: WheelFile,
+        whl_full_path: str | Path,
```
danyeaw, line 84: Nice!

tests/test_conda_meta_json_filename.py
```diff
@@ -9,26 +9,80 @@
...
-    if normalized != "py3-none-any" and normalized.endswith("-none-any"):
-        normalized = "py3-none-any"
-    assert normalized == expected
+def test_extract_whl_index_json_build_from_wheel_metadata(
+    tmp_path: Path,
+    tags: list[str],
+    wheel_build: str | None,
+    expected_build: str,
+):
+    from conda_pypi.package_extractors.whl import extract_whl_as_conda_pkg
```
danyeaw: Nit, imports should be at the top

danyeaw (commented):
Thanks Ryan, this is almost there, just a couple of additional comments to optimize test performance.

tests/test_conda_meta_json_filename.py
ryanskeith: I will move the parsing logic to a sub function and make the tests centered around assure parsing rules are followed.

tests/test_conda_meta_json_filename.py
ryanskeith: I was following the other older test as a guide and keeped it in.  I will move it to the top.

ryanskeith:
pre-commit.ci autofix

tests/test_conda_meta_json_filename.py
```diff
@@ -1,48 +1,121 @@
...
+    ("tags", "wheel_build", "expected_build"),
+    [
+        (["py310-none-any", "py3-none-any"], None, "py3_none_any_0"),
+        (["py38-none-any"], None, "py3_none_any_0"),
+        (["py2-none-any", "py3-none-any"], None, "py3_none_any_0"),
+        (["py3-none-any"], "1", "py3_none_any_1"),
+        (["cp312-cp312-win_amd64"], None, "cp312_cp312_win_amd64_0"),
+    ],
+)
+def test_create_build_string_from_wheel_meta_and_filename_(
```
danyeaw:
```suggestion
def test_create_build_string_from_wheel_meta_and_filename(
```

danyeaw (approved):
Thanks Ryan! I really like the updated approach to use unit tests 👍

PR #431, opened 2026-06-23
Skip CodSpeed benchmarks on fork repositories

danyeaw:
## Summary
- Skip the `linux-benchmarks` job on fork repositories
- Fork workflows cannot access the upstream `CODSPEED_TOKEN` secret, so CodSpeed upload fails with `401 Unauthorized`


ryanskeith (approved):
LGTM.

PR #433, opened 2026-06-25
add `wheels_local_channel` fixture

### LINKED ISSUE #427: Add offline wheel channel test fixture using `conda pypi index`
### Checklist

### What is the idea?

Add a pytest fixture that builds and serves a fully offline conda wheel channel from tests/pypi_local_index/ using conda pypi index.

### Why is this needed?

We have PyPI-style and conda-channel test fixtures today, but neither gives us a local wheel channel where repodata and wheel downloads are both offline. This would be a nice additional end-to-end test set up following #278.

### What should happen?

- Add a wheel_local_channel fixture: copy test wheels, index with conda pypi index --base-url, serve over HTTP
- Add tests for repodata (v3.whl, local URLs) and installing demo-package with Rattler

### Additional Context

- Reuse wheels already in tests/pypi_local_index/

ForgottenProgramme:
### Description
resolves #427 
Adds a fixture that yields an wheels local channel with packages from `tests/pypi_local_index/`.

### Checklist - did you ...


tests/test_wheels_local_channel.py
```diff
@@ -0,0 +1,34 @@
...
+):
+    """
+    Test that demo-package can be installed from the local wheel channel using Rattler.
+    """
+    with tmp_env(
+        "demo-package",
+        "--channel",
+        wheels_local_channel,
+    ) as prefix:
+        assert (prefix / "conda-meta").is_dir()
```
danyeaw, line 43: This check here is a bit weak since it is only checking that `conda-meta` exists, but it doesn't check that we actually installed packages successfully from the local wheels channel.

I think using a pattern like the one here would be more robust:
https://github.com/conda/conda-pypi/blob/f7719c9831d28438f389d4ce54e95497bb759983/tests/cli/test_install.py#L322

This uses a 2 step install to create an environment with the `tmp_env` and then install to it with `conda_cli`, and then check that the package from the local channel exists in the JSON.

tests/test_wheels_local_channel.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_wheels_local_channel_urls_are_local(wheels_local_channel):
+    """All wheel URLs in repodata point back to the local server."""
+    data = requests.get(f"{wheels_local_channel}/noarch/repodata.json").json()
+    for name, entry in data["v3"]["whl"].items():
+        assert entry["url"].startswith(wheels_local_channel), (
+            f"{name} has non-local URL: {entry['url']}"
+        )
+
```
danyeaw, line 21:
```suggestion
        assert entry["url"].startswith(wheels_local_channel), (
            f"{name} has non-local URL: {entry['url']}"
        )
        assert "files.pythonhosted.org" not in entry["url"]

```

danyeaw (commented):
Hi @ForgottenProgramme, this looks great! Just a couple of comments to improve the test asserts to make sure that the packages are really coming from the local channel.

tests/test_wheels_local_channel.py
ForgottenProgramme: Hey, @danyeaw !
Could you please explain this assert you're recommending. 🤔 
> `  assert "files.pythonhosted.org" not in entry["url"]`

danyeaw: Thinking about this more, I don't think this is needed - a URL can't start with the local channel and have pythonhosted in it. I'll mark it as resolved.

tests/test_wheels_local_channel.py
```diff
@@ -0,0 +1,46 @@
...
+
+def test_install_demo_package_from_wheels_local_channel(
+    wheels_local_channel,
+    with_rattler_solver,
+    tmp_env: TmpEnvFixture,
+    conda_local_channel,
```
danyeaw:
```suggestion
```

`conda_local_channel` is a test fixture for remote PyPI wheels only, so I don't think this is needed.

danyeaw (approved):
Hi @ForgottenProgramme, this looks great! Thanks so much for the implementation, just one small nit before we merge.

tests/test_wheels_local_channel.py
ForgottenProgramme: ahh, this is stale.

PR #443, opened 2026-07-02
Fix beta tip not visible at default verbosity

### LINKED ISSUE #442: Beta tip is invisible at default verbosity level
### Checklist

### What happened?

The conda-pypi beta tip (conda_pypi_pip_warning, default true) does not appear after conda create / conda install in a non-base env with pip, unless you pass -vv.

```
conda create -y -n test-pypi-tip python pip      # no tip
conda create -y -vv -n test-pypi-tip2 python pip # tip appears
```
Expected: Tip shown at normal verbosity.

Actual: Nothing shown unless log level is INFO or higher.

Cause: #387 changed notify_externally_managed_future() from logger.warning() to logger.info(). Conda's default verbosity maps to WARNING, so the message is filtered out.

Tests: tests/test_plugins.py mocks the logger and only checks that info() was called, not that users see output.

### Additional Context

Related: #387
Reproduced on conda 26.5.2, conda-pypi 0.10.1
Likely fix: use logger.warning() again (or conda's user-facing output helpers) and add a CLI integration test at default verbosity

danyeaw:
### Description
Fixes #442 by restoring the visibility level of the tip from INFO to WARNING.

### Checklist - did you ...

### DIFF
--- a/conda_pypi/main.py
+++ b/conda_pypi/main.py
@@ -153,7 +153,7 @@ def notify_externally_managed_future(command: str):
         return
 
     if context.plugins.conda_pypi_pip_warning:
-        logger.info(
+        logger.warning(
             "\n"
             "  Did you know? You can install many PyPI packages with conda\n"
             "  using the conda-pypi beta. Get started:\n"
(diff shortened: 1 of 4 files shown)


ryanskeith (approved):
This makes it visible again.

jezdez (approved):

PR #444, opened 2026-07-02
Use `md` stage for more durable wheel metadata in conda-index database

### LINKED ISSUE #403: Switch conda-pypi's indexing to use conda-index's metadata only pipeline
### What is the idea?
Move PyPI-metadata-only wheel indexing (store_pypi_metadata, test channel regen) to conda-index’s md stage.

### Why is this needed?
store_pypi_metadata registers virtual wheels in the fs stage even though no files exist on disk. That can cause spurious re-extraction #340. conda-index 0.12.0 added store_md_state for this case ([conda-index#284](https://github.com/conda/conda-index/pull/284)).

For wheels, `store_fs_state` would no longer be needed since we don't need conda-index to do extra work with the files.

### What should happen?

- Bump conda-index to >=0.12.0
- Change store_pypi_metadata to use store_md_state; add include_stages: ["md"] in PyPI-metadata-only callers
- Optionally add a shared helper so both paths reuse validation but pick the right stage
- Regenerate tests/conda_local_channel repodata and confirm tests pass

### Additional Context

dholth:
### Description

Basic usage implemented so far. When using the old technique the conda-index command would erase all of the wheel metadata if it touched the same database while listing *.conda, *.tar.bz2 from the filesystem. The `md` stage is a place to keep filenames that isn't touched by the filesystem. The caller can skip the parts of conda index that are about listing the filesystem, and creating a todo list of everything missing from "indexed" compared to the filesystem; add all metadata itself; and go straight to output. 

### Checklist - did you ...


danyeaw (approved):
Hey @dholth, thanks! I pushed a couple of changes to:
1. Bump the minimum conda-index version to 0.12.0 (that's where the md stage was added)
2. Update the local test channel repodata generation to use the md stage

tests/conda_local_channel/generate_noarch_wheel_repodata.py
```diff
@@ -54,11 +59,20 @@ def cache_repodata_entry(cache: BaseCondaIndexCache, name: str, version: str):
...
 
+    md_entries = []
     for pkg_tuple in repodata_packages:
-        cache_repodata_entry(cache, pkg_tuple[0], pkg_tuple[1])
+        entry = cache_repodata_entry(cache, pkg_tuple[0], pkg_tuple[1])
+        if entry is not None:
+            md_entries.append(entry)
+
+    if md_entries:
+        cache.store_stat_state("md", md_entries)
```
dholth: I would call this unconditionally, since "no md entries" is also a valid state. Otherwise if you delete all the packages and re-index, you'd see all the packages.

https://github.com/conda/conda-index/blob/main/conda_index/index/sqlitecache.py#L306

PR #446, opened 2026-07-08
Implement delayed imports

### LINKED ISSUE #429: Defer heavy imports to improve performance
### What is the idea?
Defer heavy third-party imports (unearth, build, installer, conda_index, and the bundled name-mapping JSON) until the code path that actually needs them runs, instead of loading them when the conda-pypi plugin registers.

### Why is this needed?
Today, importing the plugin pulls in the full conda pypi stack — including unearth, which is only used by conda pypi install. That happens on most conda CLI invocations (conda list, conda install, etc.), not just when someone runs a pypi command. This adds unnecessary startup cost for every conda user who has conda-pypi installed.

Also there are cases like #419, where conda-pypi is preventing executing conda commands due to the number of imports.

This aligns with conda core’s Track A startup work https://github.com/conda/conda/issues/15867, especially the deferred-import pattern already used in conda itself (A4). Upcoming core changes like deferred plugin discovery (A2/A3) will help, but conda-pypi still needs to defer its own imports once the plugin module is loaded.

### What should happen?

- Keep plugin.py lightweight at import time by moving conda-pypi imports inside hook implementations.
- Lazy-load subcommand modules in the CLI router by moving heavy imports from subcommand module scope into execute() (install, convert, and index only pull in what they use).
- Defer imports in hook callbacks (wheel extractor, doctor health check) and lazy-load the grayskull name-mapping JSON on first use.
- Add subprocess guard tests (same pattern as conda’s deferred-import tests) to prevent heavy modules from creeping back to import time.
- Expected outcome: conda list and similar commands no longer pay the cost of unearth/build/installer

### Additional Context

Related conda core work: https://github.com/conda/conda/issues/15867 and https://github.com/conda/conda/pull/15868.

ForgottenProgramme:
resolves #429 

### Description

- [x] Keep plugin.py lightweight at import time by moving conda-pypi imports inside hook implementations.
- [x] Lazy-load subcommand modules in the CLI router by moving heavy imports from subcommand module scope into execute() (install, convert, and index only pull in what they use).
  - [x] install
  - [x] convert
  - [x] index
- [x] Defer imports in hook callbacks (wheel extractor, doctor health check) and lazy-load the grayskull name-mapping JSON on first use.
- [x] Add subprocess guard tests (same pattern as conda’s deferred-import tests) to prevent heavy modules from creeping back to import time. Considering ["unearth", "build", "installer", "conda_index"] as heavy modules. 
- [ ] Expected outcome: conda list and similar commands no longer pay the cost of unearth/build/installer

### Checklist - did you ...

### DIFF
--- a/conda_pypi/cli/install.py
+++ b/conda_pypi/cli/install.py
@@ -107,14 +107,13 @@ def execute(args: Namespace) -> int:
     from conda.exceptions import ArgumentError
     from conda.models.match_spec import MatchSpec
     from packaging.requirements import InvalidRequirement, Requirement
-    
+
     from conda_pypi import build, convert_tree, installer
     from conda_pypi.downloader import get_package_finder
     from conda_pypi.main import run_conda_install
     from conda_pypi.markers import dependency_extras_suffix
     from conda_pypi.translate import pypi_to_conda_name, remap_match_spec_name
     from conda_pypi.utils import get_prefix
-
 
     editable_projects = args.editable or ()
     if isinstance(editable_projects, str):
--- a/conda_pypi/cli/install.py
+++ b/conda_pypi/cli/install.py
@@ -3,19 +3,7 @@ from argparse import SUPPRESS, Namespace, _SubParsersAction
 from pathlib import Path
 
 from conda.auxlib.ish import dals
-from conda.base.context import context
-from conda.cli.common import stdout_json_success
 from conda.cli.conda_argparse import add_output_and_prompt_options
-from conda.exceptions import ArgumentError
-from conda.models.match_spec import MatchSpec
-from packaging.requirements import InvalidRequirement, Requirement
-
-from conda_pypi import build, convert_tree, installer
-from conda_pypi.downloader import get_package_finder
-from conda_pypi.main import run_conda_install
-from conda_pypi.markers import dependency_extras_suffix
-from conda_pypi.translate import pypi_to_conda_name, remap_match_spec_name
-from conda_pypi.utils import get_prefix
 
 
 def configure_parser(parser: _SubParsersAction) -> None:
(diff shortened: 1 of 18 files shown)


ForgottenProgramme:
Since install.py already short-circuits to `channels=()` when `args.ignore_channels=True`, we just set that flag in the test instead of touching `context.channels` at all.

ForgottenProgramme:
@codspeedbot explain why this is faster

conda_pypi/cli/install.py
dholth, line 101: In the github diff, it looks like this is part of configure_parser() but it is a separate function.

conda_pypi/index.py
```diff
@@ -29,7 +29,6 @@ def create_channel_index(path):
 
 def update_index(channel_index: ChannelIndex):
     channel_index.index(patch_generator=None)
```
jezdez, line 31: This drops `channel_index.update_channeldata()`, so `conda pypi index` no longer writes the top-level `channeldata.json`. `conda-index 0.11.0`'s `ChannelIndex.index()` does not call this itself; indexing the demo wheel on this PR returns successfully but leaves only `noarch/repodata*` and `index.html`, while base writes `channeldata.json`. Please restore this call or add an equivalent channeldata generation path.

jezdez (changes requested):
One inline note from review, the channeldata regression is a blocker.

conda_pypi/index.py
dholth: I really dislike channeldata, will it ever be used? For example, anaconda.org doesn't support channeldata.

On the other hand, we expect conda-pypi to run on much smaller channels than `conda-forge`

ForgottenProgramme: Ah! I was scratching my head wondering why I removed that line. 
Took me a moment to realise that this is a left-over from [Daniel's commit](https://github.com/conda/conda-pypi/pull/446/changes/114d832497d7b8f6790d45772bdb3750c0a83956#diff-e7c2857b3948357eab87b8ab51ba9de18d1e3696bb97298dc49ff4e5e430b30c). Somehow not all the changes pulled into my PR when Daniel force-pushed the previous revision.

ForgottenProgramme: Good that you caught it @jezdez!

jezdez (approved):

jezdez:
For context, I ran a rough local before/after benchmark focused on the startup path this PR targets.

Method: same Pixi Python env, temp worktrees, fresh subprocess per run, median wall time.

| Scenario | Base | PR | Approx change |
|---|---:|---:|---:|
| `import conda_pypi.plugin` | 225.8 ms | 131.3 ms | -94.5 ms, ~42% faster |
| `conda_settings()` hook | 238.8 ms | 121.3 ms | -117.5 ms, ~49% faster |
| `conda_subcommands()` hook | 254.0 ms | 143.0 ms | -111.0 ms, ~44% faster |
| `python -m conda list --help` | 291.1 ms | 201.5 ms | -89.6 ms, ~31% faster |

After `import conda_pypi.plugin`, base had loaded `unearth`, `build`, `installer`, and `conda_index`; this PR had not loaded any of those yet.

So locally this saves roughly 90-115 ms on cold startup/plugin-registration paths.

PR #463, opened 2026-07-21
Mark `conda pypi install` for deprecation

### LINKED ISSUE #435: Deprecate conda pypi install in favor of conda install with the conda-pypi channel
### What is the idea?

Deprecate and eventually remove `conda pypi install` for package installs. Users should use `conda install <package>` with the conda-pypi channel and Rattler solver instead.

Out of scope: -e (separate issue), conda pypi convert, conda pypi index.

### Why is this needed?
`conda pypi install` was a great stepping stone to get to our `conda install` experience.

The `conda pypi install` conversion path installs into a local `file://…/conda-pypi` channel, which confuses users (see #422) and adds additional support burden (see https://github.com/conda/conda/issues/16279). Our docs already recommend channel-native install as the basic path and we should make it the only path.

### What should happen?

- Add pending deprecation to conda pypi install <packages> (not -e): use conda install with the conda-pypi channel instead.
- Update the docs to warn that it is deprecated
- Later: remove non-editable install logic from conda_pypi/cli/install.py.

ForgottenProgramme:
### Description
resolves #435 

Purposefully skips the `editable` install logic from deprecation marking per the instructions in the linked issue ticket. 

### Checklist - did you ...


tests/cli/test_install.py
```diff
@@ -215,32 +215,34 @@ def test_install_editable_installs_multiple_projects_in_order(
...
-            "--yes",
-            "install",
-            "--ignore-channels",
-            "--prefix",
-            prefix,
-            "--index-url",
-            pypi_local_index,
-            "demo-package",
-        )
+        with pytest.deprecated_call():
```
jezdez: `pytest.deprecated_call()` accepts only `DeprecationWarning` and `PendingDeprecationWarning`. The implementation emits `FutureWarning` starting with conda 27.3, so each new `deprecated_call()` assertion will fail when the deprecation becomes active. Please assert `(PendingDeprecationWarning, FutureWarning)` and match the `conda pypi install` warning text.

jezdez (changes requested):
The `27.3` and `27.9` dates follow CEP 9. Please fix the warning assertion described inline before merging.

Issue #435 also requires a documentation warning. This PR does not update the documentation, and the current docs still recommend the non-editable `conda pypi install` path. Please update the relevant docs.

The checklist marks the news fragment as complete, but this PR does not add one. Please add the fragment or correct the checklist.

tests/cli/test_install.py
ForgottenProgramme: I don't think that's correct. `FutureWarning` is also covered. 
https://docs.pytest.org/en/stable/reference/reference.html#pytest.deprecated_call

docs/features.md
```diff
@@ -57,6 +57,12 @@ main subcommands that handle different aspects of PyPI integration.
 
 #### `conda pypi install`
 
+:::{admonition} `conda pypi install` is pending deprecation
+:class: warning
```
ForgottenProgramme, line 61: Maybe :class: warning is too extreme for this? 🤷🏼

tests/cli/test_install.py
jezdez: You are right that pytest 8 and newer cover `FutureWarning`. This repository pins pytest 7.4.3, where it is not covered. Could you upgrade the pin to `pytest = "9.0.*"` and regenerate `pixi.lock`? With that upgrade, `deprecated_call()` can stay.

Please also match the intended warning so an unrelated deprecation cannot satisfy these assertions, and apply the same change to the other four contexts.

```suggestion
        with pytest.deprecated_call(
            match=r"`conda pypi install` for package installs"
        ):
```

jezdez (approved):

jezdez (approved):

PR #476, opened 2026-07-30
Add pixi task to add news file via CLI

ForgottenProgramme:
### Description

Add `news/add_news.py` and a `pixi run news` task so contributors can create news fragments from `news/TEMPLATE` without copying by hand.

Examples:

```bash
# create a news file with just the TEMPLATE copied into it
pixi run news 476-add-news-cli 

# create a news file with the "Enhancements" section populated with the provided `message`. 
pixi run news 476-add-news-cli --section "Enhancements" --message "Add pixi news CLI to create news fragments" 
```

### Checklist - did you ...

### DIFF
--- a/news/476-add-news-cli
+++ b/news/476-add-news-cli
@@ -1,6 +1,6 @@
 ### Enhancements
 
-* Add `pixi run news` task to create news fragments via CLI
+* Add `pixi run news` task to create news fragments via CLI. (#476)
 
 ### Bug fixes
 
--- a/news/476-add-news-cli
+++ b/news/476-add-news-cli
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Add `pixi run news` task to create news fragments via CLI
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/news/add_news.py
+++ b/news/add_news.py
@@ -1,0 +1,95 @@
+"""Script to generate news files"""
+
+import argparse
+from pathlib import Path
+
+SECTIONS_LIST = ["Bug fixes", "Docs", "Enhancements", "Deprecations", "Other"]
+
+
+def has_correct_format(filename: str):
+    """New file name format must be followed"""
+    if filename.endswith(".md"):
+        return False
+    if "-" not in filename:
+        return False
+    pr_number, _ = filename.split("-", 1)
+    return pr_number.isdigit()
+
+
+def news_cli():
+    """Logic for news file generator"""
+
+    # configure parser
+    parser = argparse.ArgumentParser(
+        prog="News File Generator",
+        description="Generate a news file with or without the news snippet, using the TEMPLATE in news/",
+    )
+
+    parser.add_argument("filename", help="Name of the news file")
+    parser.add_argument(
+        "--section", help="Section in the news template to be populated with the news snippet"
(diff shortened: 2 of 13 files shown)


scripts/add_news.py
```diff
@@ -0,0 +1,95 @@
+"""Script to generate news files"""
```
danyeaw, line 1: Rever removes everything in the news folder during a release. We would either need to move this script to scripts/ or tools/, or update the $CHANGELOG_IGNORE in `rever.xsh` upstream in infrastructure.

scripts/add_news.py
```diff
@@ -0,0 +1,95 @@
...
+        return 1
+
+    # Populate section with the provided news snippet
+    if args.section:
+        lines = template_content.splitlines()
+
+        found = False
+        for i, line in enumerate(lines):
+            if line.lower() == f"### {(args.section).lower()}":
+                lines[i + 2] = f"* {args.message}"
```
danyeaw, line 76: This is theoretically brittle if TEMPLATE spacing changes. However, I think it is fine since I wouldn't expect that we would ever change the format of the file as long as we are using rever.

pyproject.toml
```diff
@@ -69,6 +69,10 @@ description = """Set up the environment for development
                     We also need to place it in _new_ environments created by conda. We do this by
                     implementing some extra plugin hooks."""
 
+[tool.pixi.tasks.news]
+cmd="python news/add_news.py"
+description="Create a news file."
```
danyeaw:
```suggestion
cmd = "python news/add_news.py"
description = "Create a news file."
```

news/add_news.py
```diff
@@ -0,0 +1,95 @@
...
+
+import argparse
+from pathlib import Path
+
+SECTIONS_LIST = ["Bug fixes", "Docs", "Enhancements", "Deprecations", "Other"]
+
+
+def has_correct_format(filename: str):
+    """New file name format must be followed"""
+    if filename.endswith(".md"):
```
danyeaw:
```suggestion
    if "/" in filename or "\\" in filename or filename.endswith(".md"):
```

This is minor, but it might also be good to reject path separators from filenames.

news/476-add-news-cli
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add `pixi run news` task to create news fragments via CLI
```
danyeaw:
```suggestion
* Add `pixi run news` task to create news fragments via CLI. (#476)
```

danyeaw (commented):
Hey @ForgottenProgramme, this is a nice command automation! Would it make sense to also add a small blurb about it in the `CONTRIBUTING.md`?

news/add_news.py
ForgottenProgramme: I think this already got covered in :
```
if not dest.resolve().is_relative_to(news_dir.resolve()):
        print("News file must stay inside the `news/` directory")
        return 1
```

danyeaw: Hey @ForgottenProgramme, this is probably an optional change. I think the `is_relative_to` prevents escaping outside of `news/`, but it wouldn't prevent `news/476-add-news/cli` if someone passed in `pixi run news 476-add-news/cli`

ForgottenProgramme: ah! I see now. You're right. 
I will add your suggested change.

danyeaw (approved):
Thanks @ForgottenProgramme!

PR #479, opened 2026-07-30
Drop legacy extras suffix from console_scripts entry points

### LINKED ISSUE #477: conda pypi convert emits unparseable entry points for console_scripts with extras (e.g. 'mod:func [extra]')
### What happened?

`conda pypi convert` copies the **extras suffix** of a `console_scripts` entry point verbatim into `info/link.json` → `noarch.entry_points`. conda's entry-point parser rejects it, so the package solves fine and then **fails at link time** with an unhelpful crash report.

Two packages on an internal wheel-converted channel hit this:

```
google-auth-oauthlib 1.4.0  →  ValueError: 'main [tool]' is not a valid Python function identifier
mcp 1.27.2                  →  ValueError: 'app [cli]'  is not a valid Python function identifier
```

Their `link.json`:

```json
// google-auth-oauthlib 1.4.0
"noarch": {
  "entry_points": ["google-oauthlib-tool = google_auth_oauthlib.tool.__main__:main [tool]"],
  "type": "python"
}

// mcp 1.27.2
"noarch": {
  "entry_points": ["mcp = mcp.cli:app [cli]"],
  "type": "python"
}
```

conda raises in [`conda/common/path/python.py`](https://github.com/conda/conda/blob/main/conda/common/path/python.py) → `parse_entry_point_def()`, which splits on `:` and validates the remainder as a Python identifier. `main [tool]` is not one.

Upstream sources, both using the legacy setuptools extras syntax:
- [mcp 1.27.2 `pyproject.toml`](https://inspector.pypi.io/project/mcp/1.27.2/packages/27/3c/347cf965d313f5d41764e7d46bea6ffe7d9ef13b983cc429b0340962a082/mcp-1.27.2.tar.gz/mcp-1.27.2/pyproject.toml#line.47)
- [google-auth-oauthlib 1.4.0 `setup.py`](https://inspector.pypi.io/project/google-auth-oauthlib/1.4.0/packages/70/18/90c7fac516e63cf2058166fce0c88c353647c677b51cc036c09c49bb5cbb/google_auth_oauthlib-1.4.0.tar.gz/google_auth_oauthlib-1.4.0/setup.py#line.54)

### Cause

`conda_pypi/translate.py` builds the string from `ep.value`, which retains the extras suffix:

```python
# conda_pypi/translate.py
console_scripts = [
    f"{ep.name} = {ep.value}"          # ep.value == "mcp.cli:app [cli]"
    for ep in distribution.entry_points
    if ep.group == "console_scripts"
]
```

Notably the **other** code path is already correct — `conda_pypi/package_extractors/whl.py::write_script()` composes from `module` and `attr` separately, so extras never appear:

```python
entry_point = f"{name} = {module}:{attr}"
```

### Suggested fix

`importlib.metadata.EntryPoint` already parses the three parts, so this is a one-line change in `translate.py`:

```python
console_scripts = [
    f"{ep.name} = {ep.module}:{ep.attr}"
    for ep in distribution.entry_points
    if ep.group == "console_scripts"
]
```

Verified on Python 3.13:

| `ep.value` | `ep.module` | `ep.attr` | `ep.extras` |
|---|---|---|---|
| `mcp.cli:app [cli]` | `mcp.cli` | `app` | `['cli']` |
| `google_auth_oauthlib.tool.__main__:main [tool]` | `google_auth_oauthlib.tool.__main__` | `main` | `['tool']` |
| `pkg.mod:func` | `pkg.mod` | `func` | `[]` |

Dropping the extras is what the spec asks of consumers. Per [PyPA entry-points](https://packaging.python.org/en/latest/specifications/entry-points/#file-format):

> Using extras for an entry point is no longer recommended. **Consumers should support parsing them from existing distributions, but may then ignore them.** New publishing tools need not support specifying extras.

conda-pypi currently neither parses nor ignores them — it passes them through to a consumer that can't accept them.

### Reproducer

```bash
conda create -p /tmp/x --override-channels \
  -c <wheel-converted-channel> \
  -c https://repo.anaconda.com/pkgs/main \
  google-auth-oauthlib=1.4.0 python=3.11
# ValueError: 'main [tool]' is not a valid Python function identifier
```

Or directly, without a channel:

```python
from importlib.metadata import EntryPoint
ep = EntryPoint(name="mcp", value="mcp.cli:app [cli]", group="console_scripts")
print(f"{ep.name} = {ep.value}")            # mcp = mcp.cli:app [cli]   <- emitted today, unparseable
print(f"{ep.name} = {ep.module}:{ep.attr}") # mcp = mcp.cli:app         <- parseable
```

### Impact

Any wheel whose `console_scripts` uses the legacy extras syntax produces a `.conda` package that cannot be installed. The failure surfaces at link time as a generic conda crash report, so it reads as a conda bug rather than a metadata problem — it took a while to trace. Transitively, anything depending on such a package also fails: `mcp` broke `anaconda-mcp` and `jupyter-ai` for us.

### Related

- #468 — extras on *dependencies* (`coverage[toml]>=7`). Adjacent but distinct: this issue is extras on *entry points*.
- #309 — hardcoded Python paths in entry-point scripts from `conda pypi convert` (same area, closed).

Found while testing channel overlap; happy to open a PR if the suggested fix looks right.

abdul-050:
Fixes #477.

### Problem

`CondaMetadata.from_distribution` built each console script from `ep.value`, which retains the legacy setuptools extras suffix:

```python
f"{ep.name} = {ep.value}"        # ep.value == "mcp.cli:app [cli]"
```

That string lands verbatim in `info/link.json` → `noarch.entry_points`. conda's `parse_entry_point_def()` then splits on `:` and validates the remainder as a Python identifier, so `app [cli]` raises — **at link time, after the solve has already succeeded**, which surfaces as a generic conda crash report rather than a metadata error.

Observed on two wheel-converted packages:

| package | entry point | error |
|---|---|---|
| `mcp` 1.27.2 | `mcp = mcp.cli:app [cli]` | `ValueError: 'app [cli]' is not a valid Python function identifier` |
| `google-auth-oauthlib` 1.4.0 | `google-oauthlib-tool = ...:main [tool]` | `ValueError: 'main [tool]' is not a valid …` |

### Fix

Use `ep.module` / `ep.attr` — `importlib.metadata.EntryPoint` already parses the three parts and exposes extras separately:

| `ep.value` | `.module` | `.attr` | `.extras` |
|---|---|---|---|
| `mcp.cli:app [cli]` | `mcp.cli` | `app` | `['cli']` |
| `pkg.mod:func` | `pkg.mod` | `func` | `[]` |

This also brings the two conversion paths into line — `package_extractors/whl.py::write_script()` was already composing from `module`/`attr`, so it never had the bug.

Dropping extras is what the spec asks of consumers, per [PyPA entry-points](https://packaging.python.org/en/latest/specifications/entry-points/#file-format):

> Using extras for an entry point is no longer recommended. **Consumers should support parsing them from existing distributions, but may then ignore them.** New publishing tools need not support specifying extras.

### Tests

- `test_console_scripts_drop_entry_point_extras` — parametrized over `[cli]`, `[tool]`, `[a,b]`, and the no-extras case (which must stay unchanged).
- `test_link_json_entry_points_are_parseable_by_conda` — feeds the emitted `link.json` through conda's own `parse_entry_point_def()`, so a regression is caught by the real parser rather than a string comparison.

Both fail on `main` and pass with this change. Full `tests/test_translate.py` is green (34 passed).

`FileDistribution.read_text` only answers for `METADATA`, so `distribution.entry_points` was always empty in tests — the `_distribution()` helper gained an optional `console_scripts` argument backed by a small subclass that also serves `entry_points.txt`.

### DIFF
--- a/conda_pypi/translate.py
+++ b/conda_pypi/translate.py
@@ -169,8 +169,13 @@ class CondaMetadata:
         # conda does support ~=3.0.0 "compatibility release" matches
         depends = [requires_python] + requirements
 
+        # Use module/attr rather than ep.value: value retains the legacy extras
+        # suffix (``mod:func [extra]``), which conda's noarch entry-point parser
+        # rejects at link time. The spec asks consumers to parse extras and allows
+        # ignoring them, which is what dropping them here does.
+        # https://packaging.python.org/en/latest/specifications/entry-points/#file-format
         console_scripts = [
-            f"{ep.name} = {ep.value}"
+            f"{ep.name} = {ep.module}:{ep.attr}"
             for ep in distribution.entry_points
             if ep.group == "console_scripts"
         ]
--- a/news/477-entry-point-extras
+++ b/news/477-entry-point-extras
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Drop the legacy extras suffix from `console_scripts` entry points during conversion (e.g. `mod:func [extra]` → `mod:func`). Previously it was copied verbatim into `info/link.json`, making the resulting package fail to install with `ValueError: '<func> [<extra>]' is not a valid Python function identifier` at link time. (#477)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/test_translate.py
+++ b/tests/test_translate.py
@@ -143,7 +143,24 @@ def test_requires_to_conda_marker_extra_and_platform():
     assert requires == []
 
 
-def _distribution(project_urls=(), description=None, requires_python=None):
+class _EntryPointDistribution(FileDistribution):
+    """FileDistribution that also serves an ``entry_points.txt``.
+
+    ``FileDistribution.read_text`` only answers for ``METADATA``, so
+    ``distribution.entry_points`` is otherwise always empty.
+    """
+
+    def __init__(self, raw_text, entry_points_txt):
+        super().__init__(raw_text)
+        self.entry_points_txt = entry_points_txt
+
+    def read_text(self, filename: str):
+        if filename == "entry_points.txt":
+            return self.entry_points_txt
+        return super().read_text(filename)
+
+
+def _distribution(project_urls=(), description=None, requires_python=None, console_scripts=None):
     """Build a FileDistribution with the given Project-URL labels."""
     header = [
         "Metadata-Version: 2.1\n",
@@ -156,7 +173,11 @@ def _distribution(project_urls=(), description=None, requires_python=None):
     if description is not None:
         header.append("Description: " + description.replace("\n", "\n        ") + "\n")
     urls = "".join(f"Project-URL: {label}, {url}\n" for label, url in project_urls)
-    return FileDistribution("".join(header) + urls + "\n")
+    metadata = "".join(header) + urls + "\n"
+    if console_scripts is None:
+        return FileDistribution(metadata)
+    body = "".join(f"{name} = {value}\n" for name, value in console_scripts.items())
+    return _EntryPointDistribution(metadata, f"[console_scripts]\n{body}")
 
 
 @pytest.mark.parametrize(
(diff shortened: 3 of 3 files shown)


danyeaw:
@conda-bot check

danyeaw (approved):
This looks great, thanks so much @abdul-050!

PR #483, opened 2026-08-03
Remove pip dependency

danyeaw:
### Description

I noticed that we were still depending on pip from some of the old `conda pip` functionality.

This PR removes the orphaned helpers that shelled out to `pip install --dry-run` / `pip install --target` for the old explicit lockfile and wheel-unpack paths functionality and removes the dependency.

Finally, I removed `importlib_resources` from the recipe, it is in the standard library for all versions of Python we support.

### Checklist - did you ...


ForgottenProgramme (commented):
Does that mean we are skipping the deprecation process for this case? 🤔

ForgottenProgramme (approved):

danyeaw:
Hey Mahe, I think in this case skipping deprecation makes sense since this is stale code not being called by any feature or test in conda-pypi. There is a small reference at https://conda.github.io/conda-pypi/modules/#conda_pypi.installer.install_pip, but this code and the dependency should have been removed when we added the `conda pypi` command.

PR #518, opened 2026-09-08
Update recipe back to noarch

danyeaw:
### Description

In https://github.com/conda-forge/conda-pypi-feedstock/pull/67, we moved the recipe back to noarch since we aren't currently using the EXTERNALLY_MANAGED file. This updates the recipe to match.

### Checklist - did you ...

### DIFF
--- a/news/518-switch-recipe-noarch
+++ b/news/518-switch-recipe-noarch
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


ryanskeith (approved):
This looks good to me as well.

A minor thing from my packager days is that it is helpful for recipe reviews if the order of the pyproject.toml matches the recipe file.  Very minor but helpful for the package reviewers when you see the opportunity.  For next time if you remember.

PR #521, opened 2026-09-14
Require preinstalled dependencies for local builds

jezdez:
### Description

Local project builds currently install missing build requirements into the target environment. Callers installing a locked environment need to reject missing requirements without resolving or installing additional packages.

Add `install_build_dependencies=False` to `build_pypa()` and `pypa_to_conda()` to require existing build dependencies for wheel and editable builds. Both the declared build-system requirements and additional backend requirements are checked using the target Python. Existing callers retain the default behavior of installing missing requirements.

This provides the dependency policy needed by [conda-workspaces#175](https://github.com/conda-incubator/conda-workspaces/pull/175). Build-backend activation follows in stacked PR #523, which remains a draft.

### Checklist - did you ...


conda_pypi/build.py
```diff
@@ -108,10 +108,13 @@ def build_pypa(
     prefix: Path,
     distribution="editable",
     yes: bool = True,
+    install_build_dependencies: bool = True,
```
soapy1, line 111: Would it be helpful to also surface this option in the cli as a flag. eg. `conda pypi convert <package> --no-install-build-dependencies`?

tests/test_build.py
```diff
@@ -1,15 +1,120 @@
...
+    )
+    if phase == "build-package":
+        missing = "build"
+
+        def cannot_check(*args: object, **kwargs: object) -> None:
+            raise dependencies.MissingDependencyError([missing])
+
+        monkeypatch.setattr(dependencies, "check_dependencies", cannot_check)
+
+    with pytest.raises(dependencies.MissingDependencyError) as exc:
```
soapy1, line 80: Is it also worth testing that this does not produce an error if `install_build_dependencies=True`?

conda_pypi/build.py
```diff
@@ -125,10 +128,14 @@ def install_missing(requirements):
             try:
                 missing = dependencies.check_dependencies(requirements, prefix=prefix)
                 if missing:
+                    if not install_build_dependencies:
+                        raise dependencies.MissingDependencyError(missing)
```
soapy1, line 132: I found that if I install all the build deps into the prefix and try to run this, I still get a Missing Dependency error. Should that be happening? I can write up a more full example if that is helpful.

PR #529, opened 2026-09-18
Add build number parsing logic and option support

### LINKED ISSUE #315: conda pypi convert: Derive build number from wheel build tag and add --build-number override
### Checklist

### What is the idea?

When converting a wheel to a conda package, derive the conda build number from the wheel filename's build tag when present, and add a `--build-number` CLI option to allow manual override.

### Why is this needed?

Wheel filenames optionally include a build tag as the third component:
`{distribution}-{version}-{build tag}-{python tag}-{abi tag}-{platform tag}.whl`
Currently conda pypi convert does not use the build tag when setting the conda build number on the converted package, meaning this information is silently lost during conversion. Additionally, there is no way to manually set the build number, which is useful when repackaging a converted wheel without changing the upstream source version.

### What should happen?

- When a wheel filename contains a build tag, the conda build number of the converted package should be derived from it
- A --build-number option should be added to conda pypi convert to allow manual override of the derived value: `conda pypi convert --build-number <int> <wheel>`
- When --build-number is not provided, the build number is derived from the build tag if present, otherwise defaults to 0
- Invalid values (negative integers, non-integers) are rejected with a clear error message

### Additional Context

- PyPA `build tag`: Must start with a digit but may have additional characters following the number
- conda `build number`: is a non-negative integer
- conda `build string`: typically `<str>_<build number>`, for conda-pypi this is `pypi_<build_number>`

ForgottenProgramme:
### Description

`conda pypi convert` always wrote `build_number` 0, so a wheel filename build tag was dropped and there was no way to bump the number when repackaging the same version.

Convert now takes the leading digits of the PEP 427 wheel filename build tag (`1dev` becomes `1`, and missing tag becomes `0`). `--build-number` overrides that. The build string stays `pypi_<n>`. The `WHEEL` file `Build:` field is not used.

Closes #315

### Checklist - did you ...


conda_pypi/cli/convert.py
```diff
@@ -1,9 +1,19 @@
-from argparse import Namespace, _SubParsersAction
+from argparse import ArgumentTypeError, Namespace, _SubParsersAction
 from pathlib import Path
 
 from conda.auxlib.ish import dals
 
 
+def non_negative_int(value: str) -> int:
```
travishathaway, line 7: I think I would prefer calling this function `postive_int`. It's good to avoid double negations. It would be nice if the error message also said something like, "build number must be a postive integer".

travishathaway (commented):
Overall this looks good. I just have two concerns:
- The validation function name could be clearer (see comments)
- We should have some test coverage for the validator to make sure it behaves as expected (e.g. catches negative values and sends a meaning error message to the user).

conda_pypi/cli/convert.py
dholth: non-negative includes zero

conda_pypi/build.py
```diff
@@ -177,10 +178,14 @@ def build_conda(
         # This is mainly for METADATA and entry_points.txt. It would be
         # straightforward to write or find a WheelDistribution() to grab these
         # files from the wheel archive directly, instead of PathDistribution():
+        if build_number is None:
+            build_number = build_number_from_wheel_filename(whl.name)
```
danyeaw, line 182:
```suggestion
            build_number = build_number_from_build_tag(parsed.build_tag)
```

Since this is reparsing the same wheel name, it would be more efficient to get the build number from the build tag?

danyeaw (approved):
Thanks @ForgottenProgramme!

conda_pypi/build.py
ForgottenProgramme: Hey, @danyeaw !
yeah, we do parse the name twice there.
`parsed` is from `installer`, so `build_tag` is something like "7ks" or None. The `build_number_from_wheel_filename` helper function goes through `packaging`, which already splits that into (7, "ks"). Plugging `parsed.build_tag` into the current helper wouldn’t quite work without changing the types. 

That wouldn’t be difficult to do, but parsing the filename twice isn’t that expensive, I guess, and keeping the filename helper simple seems like a good idea. So I’m inclined to leave it as is. But let me know if you’d still prefer that we reuse `parsed`. I can change it. :)

danyeaw: Hey @ForgottenProgramme, I don't think this is a blocker. However, if the helper changes don't have other negatives that would cause us to prefer the current implementation, than doing less parsing and providing a more performant implementation I think is preferred.

danyeaw (approved):

PR #534, opened 2026-09-21
Remove beta status from docs and tip

### LINKED ISSUE #531: Docs: Remove that conda-pypi is in public beta
### Checklist

### What happened?

For the conda 26.9.0 release, we are announcing that conda-pypi is now stable. We need to update the docs to remove mentions of conda-pypi being beta.

### Additional Context

_No response_

bingliscodes:
### Description

Closes #531.

Removes beta framing from the docs and from the pip tip. Three of the five doc mentions needed more than a wording change, so flagging those plus two other judgment calls:

**1. `conda search` troubleshooting entry — reframed, not removed.** Sharded-repodata search landed in conda 26.7.0 (conda/conda#16134 via conda/conda#16149), but conda-pypi supports `conda >=26.1.0`, so users on 26.1–26.5.x can still hit the 404. I rewrote the cause as a version requirement and kept the dry-run workaround for older versions. Happy to drop the entry entirely if you'd rather the docs assume 26.7.0+.

**2. `EXTERNALLY-MANAGED` note — kept, reason changed.** Per your comment that placement isn't being restored, this now reads as intended behavior ("since doing so would break existing pip-based workflows") rather than a beta limitation. That's a paraphrase of your reasoning, so correct me if the framing is off.

**3. Post-transaction action renamed.** `conda-pypi-notify-pip-beta` → `conda-pypi-notify-pip-tip`, per renaming the tip in all strings. This is a registered action identifier rather than a display string, so it's the one change here with any behavioral risk. Easy to revert if you'd rather leave the name alone.

Two smaller notes. The tip text: your suggested wording came through as "using the conda-pypi: ." so I read it as `using conda-pypi. Get started:`. And the `:::{admonition} Beta` block in quickstart is now `Limitations` with `:class: warning` unchanged — the compiled-wheels and security-posture points aren't beta-dependent, and this matches the existing `Limitations` admonition in `features.md` that the adjacent `{ref}` already links to.

No new tests. The only behavioral change is the action rename, and `test_pip_tip_shows_once_on_new_install` already covers registration end-to-end by running a real `conda create` and asserting the tip is emitted. It passes on this branch.

Unrelated, noticed while setting up: `CONTRIBUTING.md` says `pixi run docs` / `pixi run serve`, but the tasks are `build-docs` / `serve-docs`. Happy to file that separately.

### Checklist - did you ...

- [ ] ~~Add / update necessary tests?~~ No new tests needed — see note above.


conda_pypi/main.py
```diff
@@ -158,7 +158,7 @@ def notify_conda_pypi_tip(
         logger.warning(
             "\n"
             "  Did you know? You can install many PyPI packages with conda\n"
-            "  using the conda-pypi beta. Get started:\n"
+            "  using conda-pypi. Get started:\n"
             "    https://docs.conda.io/projects/conda/en/stable/new-features.html\n"
```
danyeaw:
```suggestion
            "    https://bit.ly/4xXYt0B\n"
```

Let's use a URL shortener to link to the conda-pypi docs (and not require them to type quite as much.

danyeaw:
@conda-bot check

README.md
```diff
@@ -40,9 +40,8 @@ conda config --set solver rattler
 conda config --append channels conda-pypi
 ```
 
-During the beta, the `conda-pypi` channel might not appear in the
-Anaconda.org web UI and some commands such as `conda search` can fail because
-they request classic `repodata.json` metadata. Use `conda install` or
+The `conda-pypi` channel might not appear in the Anaconda.org web UI, which
+requests classic `repodata.json` metadata. Use `conda install` or
```
danyeaw:
```suggestion
```

Since anaconda.org support for the channel will go live in the next few days, I think we can remove the limitation here and in the Quickstart.

README.md
danyeaw, line 40:
```suggestion
For the best experience with `conda-pypi`, update to conda 26.9 or later. To update:

```bash
conda self update"
```

or if updating from older versions of conda without `conda-self` you can also run:

```bash
conda install --name base "conda>=26.9"
```

docs/quickstart.md
```diff
@@ -53,16 +53,15 @@ Note that the `conda-pypi` channel is currently name-mapped to `conda-forge`, no
 :::
 
 :::{note}
```
danyeaw: I think we can remove this note now that the anaconda.org web UI will be there in the next few days.

docs/quickstart.md
danyeaw:
```suggestion
conda self update
```

danyeaw:
```suggestion
```

danyeaw (commented):
Hey @bingliscodes, thanks so much for the contribution and for going through the steps to sign the CLA! Just a few suggestions to polish your great docs updates.

README.md
bingliscodes: Dropped the stray " at the end of "update" but otherwise incorporated into README.md

README.md
bingliscodes: Extended deletion to all 3 lines to include that whole sentence about the UI.

docs/quickstart.md
bingliscodes: Applied but left out the older Conda fallback, keeping quickstart minimal and letting README contain both paths.

danyeaw (approved):
Thanks @bingliscodes, this is such a great contribution and really helps us provide clear docs for folks starting to make use of conda-pypi!

bingliscodes:
One thing outside the reviewed files: docs/index.md:16 still says 

> "The tool is available in conda versions 26.5 and newer."

That's still accurate as an availability statement, but it's now the only place that doesn't have the 26.9 recommendation. Want me to add it there, or keep index.md scoped to availability?

